PR #18, opened 2025-04-06
Update dependencies in composer.lock

marcusquinn:


superdav42:
Was there a specific dependency that needs updated? It a good idea to run composer update but I was wondering if there was something specific that needed addressed.

marcusquinn:
TBH I can't remember, just running at speed among other things and littering you would "might helps" :)

superdav42:
I had run composer update in #23 so I'm closing this

Issue #15, opened 2025-03-26
Function _load_textdomain_just_in_time was called incorrectly


laurentC9:
Bug under wp 671, thank you to update this error.

Notice: Function _load_textdomain_just_in_time was called incorrectly. Translation loading for the wp-ultimo doma
in was triggered too early. This is usually an indicator for some code in the plugin or theme running too early. Translations should be loaded at the init action or later. 
Please see [Debugging in WordPress](https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/) for more information. (This message was added in version 6.7.0.) in /var/www/vhosts/lysite.com/httpdocs/wp-includes/functions.php on line 6114

Deprecated: La fonction Redux_Helpers::isFieldInUse est obsolète depuis la version Redux 4.0 ! Utilisez Redux_Helpers::is_field_in_use( $parent, $field ) à la place. in /var/www/vhosts/mysite.com/httpdocs/wp-includes/functions.php on line 6114

superdav42:
Fixed in 2.4.0

superdav42:
It'll be released tomorrow

PR #19, opened 2025-04-06
Fix issue #15: Load textdomain at init action to prevent 'too early' …

marcusquinn:
…warnings in WP 6.7+

https://github.com/superdav42/wp-multisite-waas/issues/15

### DIFF
--- a/inc/class-wp-ultimo.php
+++ b/inc/class-wp-ultimo.php
@@ -231,9 +231,11 @@ final class WP_Ultimo {
 	 */
 	public function setup_textdomain(): void {
 		/*
-		 * Loads the translation files.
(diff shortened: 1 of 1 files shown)


superdav42:
Am afraid fixing this is much more involved than this change. I tried it like this at first and it broke many things, like the Ajax endpoint which needed to be included earlier. I have a branch with a big refactor that addresses this issue and many others.

marcusquinn:
should still be safe to merge

superdav42:
This change is not needed and didn't really fix the original problem. Changing `load_plugin_textdomain()` to the `init` hook would stop the notices but it would also break ever call to __() that happened before init. The solution which is already implemented in main was to delay translating any string with `__()` until after the init hook. load_plugin_textdomain can still happen when the plugin loads.

Issue #10, opened 2025-02-20
Uncaught Error: Class "WP_Ultimo\Database\Sites\Site_Query" not found


techdaddies-kevin:
Plugin is network activated. Visiting the Template site that WP Ultimo created causes a Fatal error on both the front-end and back-end of the domain:

[...]

PR #20, opened 2025-04-06
Fix 10

marcusquinn:
potential fix for https://github.com/superdav42/wp-multisite-waas/issues/10

needs testing

### DIFF
--- a/inc/class-wp-ultimo.php
+++ b/inc/class-wp-ultimo.php
@@ -231,9 +231,11 @@ final class WP_Ultimo {
 	 */
 	public function setup_textdomain(): void {
 		/*
-		 * Loads the translation files.
+		 * Loads the translation files at the init action to prevent "too early" warnings in WP 6.7+
 		 */
-		load_plugin_textdomain('wp-ultimo', false, dirname((string) WP_ULTIMO_PLUGIN_BASENAME) . '/lang');
+		add_action('init', function() {
+			load_plugin_textdomain('wp-ultimo', false, dirname((string) WP_ULTIMO_PLUGIN_BASENAME) . '/lang');
+		});
 	}
 
 	/**
--- a/wp-multisite-waas.php
+++ b/wp-multisite-waas.php
@@ -67,6 +67,25 @@ require_once __DIR__ . '/vendor/woocommerce/action-scheduler/action-scheduler.ph
 require_once __DIR__ . '/inc/traits/trait-singleton.php';
 
 /**
+ * Fix for "Class 'WP_Ultimo\Database\Sites\Site_Query' not found" error
+ * Ensure all necessary database classes are loaded explicitly
+ */
+// BerlinDB core classes
+if (file_exists(__DIR__ . '/vendor/berlindb/core/src/Database/Base.php')) {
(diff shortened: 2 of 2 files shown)


superdav42:
@marcusquinn I have a huge branch I've been working on. I'll try to get it merged in Monday but your PRs will probably need rebased. Once my large refactor is in I'll try to commit things more often and use PRs since it seems I'm not a solo developer anymore. Thanks for your contributions.

marcusquinn:
Nice! And pleased to virtually meet.

Sure, feel free to reject anything, and I can do again once you've updated main.

I'm bringing some other devs to this party, too, so feel free to log more of your ideas in issues and I'll see who I can get to pick up some and help make this all we know it can be!

marcusquinn:
looks like this should still be safe to merge

inc/class-wp-ultimo.php
```diff
@@ -231,9 +231,11 @@ public function is_loaded() {
 	 */
 	public function setup_textdomain(): void {
 		/*
-		 * Loads the translation files.
+		 * Loads the translation files at the init action to prevent "too early" warnings in WP 6.7+
 		 */
-		load_plugin_textdomain('wp-ultimo', false, dirname((string) WP_ULTIMO_PLUGIN_BASENAME) . '/lang');
+		add_action('init', function() {
```
superdav42, line 240: Is this still needed?

wp-multisite-waas.php
```diff
@@ -66,6 +66,25 @@
 
 require_once __DIR__ . '/inc/traits/trait-singleton.php';
 
+/**
+ * Fix for "Class 'WP_Ultimo\Database\Sites\Site_Query' not found" error
+ * Ensure all necessary database classes are loaded explicitly
+ */
+// BerlinDB core classes
+if (file_exists(__DIR__ . '/vendor/berlindb/core/src/Database/Base.php')) {
```
superdav42: Can these be move to the Sunrise class? https://github.com/superdav42/wp-multisite-waas/blob/main/inc/class-sunrise.php#L129 most of them are already here and I would expect it to fix the problem. I don't want to put things in the main plugin file unless we absolutely have to.

marcusquinn:
Hi @superdav42,

I've updated the PR to address your feedback. Here's what I've done:

1. **Moved the Site_Query loading to the Sunrise class** as you suggested. The code is now in the `load_dependencies()` method in `inc/class-sunrise.php`. This implementation:
   - Checks if the BerlinDB files exist before loading them
   - Loads the Schema class which is needed for proper database functionality
   - Explicitly loads the sites-schema.php file which is required for the Site_Query class to work properly

2. **Kept the textdomain loading fix** that wraps the loading in an 'init' action hook. This is still needed to fix issue #15 which reports "too early" warnings in WordPress 6.7+. The fix prevents the warning message:
   ```
   Warning: load_textdomain was called with the "wp-ultimo" domain too early in the WordPress bootstrap process. This can lead to translation problems.
   ```

I've also cleaned up the implementation to make it more focused on just these two issues. The PR now only contains the necessary changes to fix issues #10 and #15.

Let me know if you need any further adjustments!

superdav42 (changes requested):
Remove the text domain changes from this PR so it's only related to the db classes missing

Issue #5, opened 2025-01-23
Domain Mapping Issue


VirgoSocials:
Hi there,

The fork works except the domain mapping no longer works for some reason, i get redirected to the main site not the website that is mapped.

Thanks

yward:
Make sure that wp-config has the following line:
define('SUNRISE', true);

PR #22, opened 2025-04-06
Fix issue #5: Add SUNRISE constant check and notification for domain …

marcusquinn:
…mapping

Proposed improvement for https://github.com/superdav42/wp-multisite-waas/issues/5

Needs testing, if you'd be so kind @virgosocials @yward

Cursor + Claude Sonnet 3.7 vibe patching, here for inspiration in getting this project stable and rolling again.


yward:
Note to self while testing:
The addon checks if sunrise==true, however there is another way to add it to wp-config,which is define( 'SUNRISE', '1' );  not sure if this will also be covered during the check, i.e. if 1 will be interpreted as true if somebody already has it added that way to avoid duplicate entry.

marcusquinn:
@yward Thank you for pointing this out! You're absolutely right that we need to handle different ways of defining the SUNRISE constant.

I've created an improved version of the PR that addresses this issue: https://github.com/superdav42/wp-multisite-waas/pull/45
1. Updated the SUNRISE check to handle multiple formats: `true`, `1`, `'1'`, and `'true'` (case-insensitive)
2. Added documentation in the readme.txt to mention that both `define('SUNRISE', true);` and `define('SUNRISE', '1');` are valid options
3. Updated the AJAX handler to also check for these different formats

The changes are in a new branch: https://github.com/marcusquinn/wp-multisite-waas/tree/fix-5-improved

I've create a separate PR for these improvements to keep the original PR clean and focused.

superdav42:
@marcusquinn Actually it'd be better for such incremental improvements to be added into the original PR. I'll just close this one in favor of https://github.com/superdav42/wp-multisite-waas/pull/45

PR #23, opened 2025-04-07
Prep Plugin for release on WordPress.org

superdav42:
Escape everything that should be escaped.
Add nonce checks where needed.
Sanitize all inputs.
Apply Code style changes across the codebase.
Correct many deprecation notices.
Optimize load order of many filters.


views/wizards/setup/default.php
```diff
@@ -6,54 +6,37 @@
...
+    <p class="wu-text-lg wu-text-gray-600 wu-mt-4 wu-mb-0">
+        <?php echo esc_html($description); ?>
+    </p>
 <?php endif; ?>
 
 <div class="wu-bg-white wu-p-4 wu--mx-5">
-
-	<?php echo $content; ?>
-
+    <?php echo wp_kses_post($content); ?>
```
superdav42:
```suggestion
    <?php echo $content; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
```

superdav42 (commented):
correct

PR #26, opened 2025-04-07
Add GitHub workflow for automatic plugin releases

marcusquinn:


superdav42:
I just committed a build step to #23. Should be as easy as `npm run build` to create the archive and now it's under 10mb. But I'm interested it making it part of this git workflow. The workflow just needs updated to run `npm run build`. It will makepot, minify css and js and other things that this does not do now.  When does this github workflow trigger to actually create the release?

PR #27, opened 2025-04-07
Add a nice Readme.md and cleanup readme.txt for wordpress.org

marcusquinn:


marcusquinn:
Yeah. Although each serves kinda different purposed, since readme.txt us mostly WP User guidance, and README.md is more for the GitHub collaborators, so doesn't need to be packaged for installer .zip.

PR #37, opened 2025-04-15
Fix 36 whitelabel css

### LINKED ISSUE #36: whitelabel.min.css 404
A dependency seems to be missing from the file in this directory: wp-content/plugins/wp-multisite-waas/assets/css/whitelabel.min.css

Image

marcusquinn:
# Fix for Issue #36: whitelabel.min.css 404

## Description
This PR fixes the 404 error for the `whitelabel.min.css` file by modifying the `wu_get_asset()` function to check if the minified file exists before trying to use it. If the minified file doesn't exist, it falls back to the non-minified version.

## Changes Made
1. Modified the `wu_get_asset()` function in `inc/functions/assets.php` to:
   - Check if the minified file exists before trying to use it
   - Fall back to the non-minified version if the minified version doesn't exist
2. Regenerated the minified CSS file to ensure it exists

## Testing
1. Disable SCRIPT_DEBUG in your WordPress configuration
2. Visit the admin area with the whitelabel feature enabled
3. Check the browser console - there should be no 404 errors for the whitelabel.min.css file

## Screenshots
Before:
Before fix - 404 error

After:
No 404 errors in the console.

## Related Issues
Fixes #36

### DIFF
--- a/inc/class-sunrise.php
+++ b/inc/class-sunrise.php
@@ -137,8 +137,10 @@ class Sunrise {
 		require_once __DIR__ . '/class-settings.php';
 		require_once __DIR__ . '/limits/class-plugin-limits.php';
 		require_once __DIR__ . '/limits/class-theme-limits.php';
-		require_once __DIR__ . '/limits/class-theme-limits.php';
 		require_once __DIR__ . '/models/class-membership.php';
+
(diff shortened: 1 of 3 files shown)


superdav42:
I don't think I want this change. I added a build script which will always generate the min files so we should be ok

PR #38, opened 2025-04-15
Perf opt 1 conditional asset loading

marcusquinn:
# Performance Optimization: Conditional Script and Style Loading

## Description
This PR implements conditional script and style loading to reduce unnecessary asset loading in the admin area. Instead of loading all scripts and styles on every admin page, we now only load them on pages where they're actually needed.

## Changes Made
1. Modified `inc/class-scripts.php` to conditionally load scripts and styles based on the current admin page
2. Added helper methods to determine which pages need which scripts and styles
3. Created specific page groups for different types of assets:
   - List pages (that need list table scripts)
   - Vue pages (that need Vue.js)
   - Modal pages (that need wubox)
   - Flag pages (that need country flag styles)

## Benefits
- Reduced page load time in the admin area
- Decreased memory usage
- Improved overall performance
- Reduced potential for script conflicts

## Testing
1. Visit different WP Ultimo admin pages and verify that functionality works correctly
2. Check browser network tab to confirm that only necessary scripts and styles are loaded
3. Verify that no JavaScript errors occur on any admin pages

## Performance Impact
- Before: All scripts and styles were loaded on every admin page
- After: Scripts and styles are only loaded when needed

This optimization is particularly noticeable on admin pages that don't require complex functionality, where the number of loaded assets is significantly reduced.

### DIFF
--- a/inc/class-scripts.php
+++ b/inc/class-scripts.php
@@ -335,7 +335,27 @@ class Scripts {
 	 */
 	public function enqueue_default_admin_styles(): void {
 
-		wp_enqueue_style('wu-admin');
+		// Get current screen to conditionally load styles
+		$screen = get_current_screen();
+
+		if (!$screen) {
+			return;
+		}
+
+		// Only load styles on WP Ultimo admin pages
+		if ($this->is_wp_ultimo_admin_page($screen)) {
+			wp_enqueue_style('wu-admin');
+
+			// Load flag styles only when needed (e.g., on pages with country selection)
+			if ($this->is_wp_ultimo_flags_page($screen)) {
+				wp_enqueue_style('wu-flags');
+			}
+		}
+
+		// Load checkout styles only on checkout pages
+		if (isset($_GET['page']) && strpos($_GET['page'], 'wu-checkout') === 0) {
+			wp_enqueue_style('wu-checkout');
+		}
 	}
 
 	/**
@@ -346,7 +366,32 @@ class Scripts {
 	 */
 	public function enqueue_default_admin_scripts(): void {
 
-		wp_enqueue_script('wu-admin');
+		// Get current screen to conditionally load scripts
+		$screen = get_current_screen();
+
+		if (!$screen) {
+			return;
+		}
+
+		// Base admin script for all WP Ultimo admin pages
+		if ($this->is_wp_ultimo_admin_page($screen)) {
+			wp_enqueue_script('wu-admin');
+		}
+
+		// Only load these scripts on specific pages where they're needed
+		if ($this->is_wp_ultimo_list_page($screen)) {
+			wp_enqueue_script('wu-ajax-list-table');
+		}
+
+		// Load Vue apps only on pages that need them
+		if ($this->is_wp_ultimo_vue_page($screen)) {
+			wp_enqueue_script('wu-vue-apps');
+		}
+
+		// Load wubox only when needed
+		if ($this->is_wp_ultimo_modal_page($screen)) {
+			wp_enqueue_script('wubox');
+		}
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


inc/class-scripts.php
```diff
@@ -335,7 +335,27 @@ public function register_default_styles(): void {
...
+			wp_enqueue_style('wu-admin');
+
+			// Load flag styles only when needed (e.g., on pages with country selection)
+			if ($this->is_wp_ultimo_flags_page($screen)) {
+				wp_enqueue_style('wu-flags');
+			}
+		}
+
+		// Load checkout styles only on checkout pages
+		if (isset($_GET['page']) && strpos($_GET['page'], 'wu-checkout') === 0) {
```
superdav42, line 356: please use str_starts_with instead. the polyfill is included in vendor so it will always work even on PHP 7.4

inc/class-scripts.php
```diff
@@ -382,4 +427,116 @@ public function add_body_class_container_boxed($classes) {
...
+	 *
+	 * @since 2.0.0
+	 *
+	 * @param \WP_Screen $screen The current screen object.
+	 * @return bool
+	 */
+	private function is_wp_ultimo_admin_page($screen) {
+		// Check if the screen ID contains 'wp-ultimo' or if it's a WP Ultimo admin page
+		return (
+			strpos($screen->id, 'wp-ultimo') !== false ||
```
superdav42, line 442: use str_contains and str_starts_with, much easier to read

inc/class-scripts.php
```diff
@@ -382,4 +427,116 @@ public function add_body_class_container_boxed($classes) {
...
+			'wu-customers',
+			'wu-payments',
+			'wu-discount-codes',
+			'wu-domain-mappings',
+			'wu-sites',
+		);
+
+		return $this->is_wp_ultimo_admin_page($screen) &&
+			   isset($_GET['page']) &&
+			   in_array($_GET['page'], $list_pages);
```
superdav42, line 470:
```suggestion
			   in_array($_GET['page'], $list_pages, true);
```
add true to third param to enable strict mode.

inc/class-scripts.php
```diff
@@ -382,4 +427,116 @@ public function add_body_class_container_boxed($classes) {
...
+	 * @return bool
+	 */
+	private function is_wp_ultimo_flags_page($screen) {
+		// Pages that use country flags
+		$flags_pages = array(
+			'wu-settings',
+			'wu-add-customer',
+			'wu-edit-customer',
+			'wu-add-payment',
+			'wu-edit-payment',
```
superdav42, line 535: I'm pretty sure the dashboard uses flags in some cases too

inc/class-scripts.php
```diff
@@ -382,4 +427,116 @@ public function add_body_class_container_boxed($classes) {
...
+		);
+
+		return $this->is_wp_ultimo_admin_page($screen) &&
+			   isset($_GET['page']) &&
+			   in_array($_GET['page'], $flags_pages);
```
superdav42, line 540:
```suggestion
			   in_array($_GET['page'], $flags_pages, true);
```

inc/class-scripts.php
```diff
@@ -382,4 +427,116 @@ public function add_body_class_container_boxed($classes) {
...
+	private function is_wp_ultimo_list_page($screen) {
+		// Pages that use list tables
+		$list_pages = array(
+			'wu-memberships',
+			'wu-products',
+			'wu-customers',
+			'wu-payments',
+			'wu-discount-codes',
+			'wu-domain-mappings',
+			'wu-sites',
```
superdav42, line 465: I don't much like having the lists like these and they would have to be kept up-to-date in multiple places. 
Actually I think the way this is written it's enqueing the script twice and maybe more than needed. 
It's already happening here:
https://github.com/superdav42/wp-multisite-waas/blob/d88e50df38fa8b402fadb7da80b484aec043b020/inc/list-tables/class-base-list-table.php#L408
and the base class make it so it's always included on pages that need it.

inc/class-scripts.php
```diff
@@ -346,7 +366,32 @@ public function enqueue_default_admin_styles(): void {
...
+
+		// Load Vue apps only on pages that need them
+		if ($this->is_wp_ultimo_vue_page($screen)) {
+			wp_enqueue_script('wu-vue-apps');
+		}
+
+		// Load wubox only when needed
+		if ($this->is_wp_ultimo_modal_page($screen)) {
+			wp_enqueue_script('wubox');
+		}
```
superdav42, line 394: These are already being enqueued is the admin page classes that use them.

superdav42 (changes requested):
The only part of this PR we need is the one that only loads the scripts on styles on WP Multisite Admin pages instead of all pages. The rest is extra code that does nothing helpful.

superdav42:
closed in favor of #54

PR #42, opened 2025-04-15
Performance Optimization: Enhance cache manager with selective cache flushing...

marcusquinn:
# Performance Optimization: Enhanced Cache Manager with Selective Cache Flushing

## Description
This PR enhances the Cache Manager to support selective cache flushing, allowing for more targeted cache clearing operations. This improves performance by avoiding unnecessary cache flushes and provides more control over which caches are cleared.

## Changes Made
1. Modified `flush_known_caches()` to accept an optional array of cache types to flush
2. Added a new `get_available_cache_types()` method to retrieve all available cache types
3. Added a new `is_cache_type_available()` method to check if a specific cache type is available
4. Added a new `wp_object_cache_flush()` method to flush WordPress object cache
5. Added a new `flush_wp_ultimo_caches()` method to flush WP Ultimo specific caches
6. Added a new `clear_cache_group()` method to clear specific cache groups

## Benefits
- More granular control over cache flushing operations
- Reduced performance impact when clearing caches
- Ability to target specific caching systems instead of flushing all caches
- Better organization of WP Ultimo specific caches
- Improved developer API for cache management

## Testing
1. Test selective cache flushing by specifying specific cache types
2. Verify that WP Ultimo specific caches are properly cleared
3. Check that all existing cache flushing functionality still works correctly

## Performance Impact
- Before: All caches were flushed at once, even when only specific caches needed to be cleared
- After: Caches can be selectively flushed, reducing unnecessary cache clearing operations

This optimization is particularly effective for sites with multiple caching systems, where flushing all caches can be resource-intensive and time-consuming.

### DIFF
--- a/inc/managers/class-cache-manager.php
+++ b/inc/managers/class-cache-manager.php
@@ -27,9 +27,10 @@ class Cache_Manager {
 	 * Flush known caching plugins, offers hooks to add more plugins in the future
 	 *
 	 * @since 2.1.2
+	 * @param array $types Optional. Array of cache types to flush. If empty, all caches will be flushed.
 	 * @return void
 	 */
-	public function flush_known_caches(): void {
+	public function flush_known_caches(array $types = array()): void {
 
 		/**
 		 * Iterate through known caching plugins methods and flush them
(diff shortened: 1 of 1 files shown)


superdav42:
I'm gonna close this because I don't see a real need for it. It adds many useful methods but unless there is something that uses them it's just dead code. If you plan to add something to use the new methods include it in this PR and reopen it.

PR #46, opened 2025-04-16
Fix divi modules pro conflict

marcusquinn:
Potential fix for https://github.com/superdav42/wp-multisite-waas/issues/9#issuecomment-2808150924


superdav42:
This is a lot of code to add for problems that don't seem to be related to our plugin. The unlimited memory setting would not trigger the error reported in the issue. Did you test the plugin combination and reproduce the error?

marcusquinn:
true, could always remove it later - just threw something up for @twicealive to test really - will see if we can make it more agnostic...

PR #58, opened 2025-04-17
Fix wiki sidebar navigation and GitHub Action workflow

marcusquinn:
iterating improvements, more to come, just needed to see if this fixes the sidebar issue


marcusquinn:
Much better! At least now we have something to evolve from what we inherited.

PR #60, opened 2025-04-17
Add GitHub workflow for automatic plugin releases

marcusquinn:
# Add GitHub Workflow for Automatic Plugin Releases

## Description
This PR adds a GitHub Actions workflow that automatically builds and releases the plugin when a new tag is pushed to the repository. This simplifies the release process and ensures consistent builds.

## Changes Made
1. Added `.github/workflows/release.yml` with the release workflow configuration
2. Updated `.github/workflows/README.md` to include documentation about the release workflow

## How It Works
1. When a new tag is pushed (in the format `v*.*.*`), the workflow is triggered
2. The workflow checks out the code, sets up PHP and Node.js
3. It verifies that the version in the tag matches the version in the plugin files
4. It runs the build process using `npm run build`
5. It creates a ZIP file of the plugin
6. It creates a GitHub release with the ZIP file attached

## Usage
To create a new release:

1. Update the version number in:
   - `wp-multisite-waas.php` (the `Version:` header)
   - `readme.txt` (the `Stable tag:` field)
   - `package.json` (the `version` field)

2. Commit these changes

3. Create and push a new tag:
   ```
   git tag v1.2.3
   git push origin v1.2.3
   ```

4. The workflow will automatically create a release with the built plugin

## Note on Previous PR
This PR is a focused version of PR #26, which included many additional files and changes. Based on the discussion in PR #26, I've created this new PR with only the essential files needed for the GitHub workflow for automatic plugin releases.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,8 +67,8 @@ jobs:
           files: |
             build/wp-multisite-waas-${{ env.VERSION }}.zip
           name: WP Multisite WaaS ${{ env.VERSION }}
-          draft: false
-          prerelease: false
+          draft: true
+          prerelease: true
           body: |
             # WP Multisite WaaS ${{ env.VERSION }}
             
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,89 @@
+name: Release WordPress Plugin
+
+on:
+  push:
+    tags:
+      - 'v*.*.*'
+
+jobs:
+  build:
+    name: Build and Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,89 @@
...
+
+      - name: Create Release
+        id: create_release
+        uses: softprops/action-gh-release@v1
+        with:
+          files: |
+            build/wp-multisite-waas-${{ env.VERSION }}.zip
+          name: WP Multisite WaaS ${{ env.VERSION }}
+          draft: false
+          prerelease: false
```
superdav42:
```suggestion
          draft: true
          prerelease: true
```

superdav42 (changes requested):
Let's set it to draft and prerelease at least until I'm confident it's going to work correctly. Do you know a way to test it without merging it first?

PR #64, opened 2025-04-18
Fix domain record creation during site creation/duplication

marcusquinn:
# Fix Domain Record Creation and Add DNS Check Interval Setting

## Description
This PR addresses two related issues with domain management:

1. Fixes an issue where domain records are not being created automatically when a site is created or duplicated. This is particularly important for the Closte.com integration, as the domain record creation triggers the Closte API to create the domain and SSL certificate.

2. Adds a configurable setting for the DNS and SSL certificate checking interval. Previously, this interval was hardcoded to 5 minutes (300 seconds), which was too long for practical use during site setup and testing.

### Domain Record Creation Fix
- Added a `create_domain_record_for_site` method to the `Domain_Manager` class that creates a domain record for a site if one doesn't already exist
- Modified the `handle_site_created` method to call the new method for both subdomain and non-subdomain sites
- Added a hook to the `Site` model's `save` method to ensure domain records are created when sites are created through the WP Ultimo interface
- Added a hook to the `Domain_Manager` class to handle the `wu_initialize_site` action

### DNS Check Interval Setting
- Added a new setting field `dns_check_interval` to the domain mapping settings section
- Modified the `async_process_domain_stage` method to use the new setting value instead of the hardcoded 5 minutes
- Added validation to ensure the value is between 10 and 300 seconds
- Updated the scheduling to use seconds instead of minutes for more precise control
- Updated log messages to reflect the change from minutes to seconds

## Settings Details
- Default value: 300 seconds (5 minutes)
- Minimum value: 10 seconds
- Maximum value: 300 seconds (5 minutes)
- Setting is accessible in the Domain Mapping settings section

## Testing
These changes have been tested to ensure that:
- Domain records are created automatically when a site is created or duplicated
- The Closte.com integration is triggered when a domain record is created
- The DNS and SSL certificate checks use the configured interval
- The interval is properly validated to be within the allowed range

## Related Issues
- Fixes the issue where site creation/duplication doesn't create domain records
- Addresses the request for a configurable DNS checking interval setting


superdav42 (approved):
I removed some redundant code. This is G2G now.

PR #75, opened 2025-04-28
Use Runcloud v3 api

### LINKED ISSUE #48: RunCloud Integration probably broken
I was just notified by RunCloud’s support team that they have moved to V3 of their API which utilizes bearer token for authentication - so their platform is apparently no longer natively compatible with Ultimo. 😭 [https://runcloud.io/docs/api/v3/#authentication](https://runcloud.io/docs/api/v3/#authentication)

 Might still work for those with existing API 2 tokens, but won’t work for new users via API 3.

superdav42:
fixes #48 
but generated by AI. Needs reviewed and tested.

### DIFF
--- a/inc/integrations/host-providers/class-runcloud-host-provider.php
+++ b/inc/integrations/host-providers/class-runcloud-host-provider.php
@@ -9,7 +9,8 @@
 
 namespace WP_Ultimo\Integrations\Host_Providers;
 
-use Psr\Log\LogLevel;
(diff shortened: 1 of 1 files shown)


superdav42:
More work is needed. It should continue to use V2 of the api if it's using the old config. New users should be prompted to use the bearer token but we don't want to break existing, working sites and force them to re-configure runcloud if v2 is still functional.

superdav42:
This is ready to test now

PR #103, opened 2025-07-10
Fix domain mapping issue in sunrise.php

tamkendigital:
Fix domain mapping not working because use old name


superdav42:
I remember finding this problem and fixing it but I guess somehow the commit got lost. Thanks for finding it!

PR #119, opened 2025-07-25
fix build process on windows

pgroot91:
Was unable to get things going on Windows. This should also be tested on Linux and Mac users.


pgroot91:
Wondering why the makepot is part both of the build and build:dev process? 

I needed to introduce `wp-cli/wp-cli-bundle` as a dev package to make the build:dev and makepot working. During the build process dev packages are not available on windows so will fail so not completely sure what to do with that.

superdav42:
Makepot doesn't need to be in the build:dev. That was just an oversight.

I was wondering what part of the original script wasn't working on Windows? This changes look fine but I'm reluctant to make the build process unnecessarily complex.

pgroot91:
@superdav42 almost all steps didn't work because they where all writen unix style. Nothing has really changed other than making this build process work cross platform. Also feedback was scares from all the errors so here and there also added feedback in the console.

Issue #84, opened 2025-05-19
support more currencies


superdav42:
at least everyone supported by stripe. We have a request for 
Name : Tanzanian Shillings
Currency code : TZS
currency symbol : TSh

PR #122, opened 2025-08-02
Update currencies

pgroot91:
**Removed the following currencies**

- Antillean Guilder (ANL), replaced by Caribbean Guilder (XCG) since this is the new currency per June 2025.

**Added the following currencies**

- Belarusian Ruble
- Myanmar Kyat
- Caribbean Guilder

**Updated the following currencies iso codes**

- Mauritanian Ouguiya
- Sierra Leonean Leone

This is updated based on the [stripe documentation](https://docs.stripe.com/currencies#presentment-currencies)

#84


inc/functions/currency.php
```diff
@@ -261,6 +265,12 @@ function wu_get_currency_symbol($currency = '') {
 		case 'KRW':
 			$currency_symbol = '₩';
 			break;
+		case 'MMK':
+			$currency_symbol = 'K, Ks';
```
superdav42:
```suggestion
			$currency_symbol = 'K';
```
I think the symbol should just be 'K'  according to [wikipedia](https://en.wikipedia.org/wiki/Malawian_kwacha). Declaring it this way would display the price like "100 K, Ks" which is probably not what we want.

superdav42 (changes requested):
Thank for this. I had done a quick and dirty review of the currencies before 2.4.0 was released and added several. I found and old dump of symbols and names and codes that was a little out of date, so this is much better. Just update MMK symbol and it's g2g.

superdav42 (approved):

PR #124, opened 2025-08-03
Add phpMyAdmin and Mailpit in docker workflow

pgroot91:
1. Make available phpMyAdmin and make sure both the development and test database are accessable. ✅ 
2. Add mailpit for testing emails and could also be used during development. ✅ 
3. Added a simple test proving the implementation works with mailpit api for email testing. ✅ 
4. Updated the pipeline and included mailpit. ✅


superdav42 (approved):
This is great! Thanks for updating the readme and everything. The only nitpick is the indention was changed for the entire composer.json file which caused conflicts. I undid the indention changes but preserved the additions. 

Thanks!

pgroot91:
@superdav42 so what rules are followed and how to set it up properly... i don't see anything in the project. 

Maybe we should add linters for the project if we want to follow something.

PR #138, opened 2025-08-05
Fix line items deserialization error in payment admin interface

vuckro:
## Problem

Line items in the payment admin interface were causing PHP errors due to improper deserialization from database storage. When viewing payments with line items in the admin panel, the system would attempt to use array data as `Line_Item` objects.

### Symptoms
- PHP fatal errors when viewing payments in admin: `Call to a member function on array`
- Line items not displaying correctly in payment details
- Admin interface broken for payments containing line items
- Error logs showing type mismatch issues

## Root Cause

**File:** `inc/models/class-payment.php` (method `get_line_items()`, line ~598-607)

The `get_line_items()` method retrieved line items from database metadata but didn't properly convert serialized array data back to `Line_Item` objects:

```php
public function get_line_items(): array {
    if (null === $this->line_items) {
        $line_items = (array) $this->get_meta('wu_line_items');
        $this->line_items = array_filter($line_items); // ❌ Arrays used as objects
    }
    return (array) $this->line_items;
}
```

## Solution

Added proper deserialization logic to convert array data back to `Line_Item` objects:

```diff
public function get_line_items(): array {
    if (null === $this->line_items) {
        $line_items = (array) $this->get_meta('wu_line_items');
        
+       // Convert arrays back to Line_Item objects
+       $line_items = array_map(function($item) {
+           if (is_array($item)) {
+               return new \WP_Ultimo\Checkout\Line_Item($item);
+           }
+           return $item;
+       }, $line_items);

        $this->line_items = array_filter($line_items);
    }
    return (array) $this->line_items;
}
```

## Technical Details

1. **Serialization Issue:** WordPress `get_meta()` returns serialized data as arrays
2. **Object Reconstruction:** Line item arrays need to be converted back to `Line_Item` objects
3. **Type Safety:** Added type checking to handle both array and object formats
4. **Backward Compatibility:** Existing `Line_Item` objects are preserved unchanged

## Benefits

- ✅ **Admin Interface:** Payment details now display correctly
- ✅ **Error Prevention:** Eliminates PHP fatal errors in admin
- ✅ **Data Integrity:** Proper object-oriented line item handling
- ✅ **Performance:** Minimal overhead with lazy loading preservation

## Testing

- [x] Verified payment admin interface displays line items correctly
- [x] Tested with existing payments containing line items
- [x] Confirmed no PHP errors in error logs
- [x] Validated line item calculations remain accurate
- [x] Tested payment creation and editing workflows

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Files Changed

- `inc/models/class-payment.php` - Added line item deserialization logic


superdav42:
I think the better solution is to prevent arrays from being serialized in the database. I believe this is happening when payments are created with the API. I just made this change https://github.com/superdav42/wp-multisite-waas/commit/14931bf2158c9c7f9b559cf33a62829ceb8b008d#diff-0e9e32e96750797b179eda709b0010fc22de54fc734fbbd82c13032b9d5e974bR621-R622 
which ensures that line items will be the line item object. So essentially it will ensure that line items are serialized correctly so they won't need to be checked when deserialized. Always better to do such checks when writing instead of reading.

PR #143, opened 2025-08-06
Update class-base-stripe-gateway.php

JoseVega:
Add hook to let other plugins change the success_url and cancel_url


JoseVega:
Hi David. Your change has a syntax error, and $subscription_data is no longer necessary as a parameter in the hook since you're trying to pass the array directly.

PR #149, opened 2025-08-07
Fix API validation: Change gateway_customer_id type to string in membership-create schema

vuckro:
###   **Fixes "Bad request - Invalid parameter(s): gateway_customer_id" error when creating memberships via POST API.**

Gateway customer IDs are typically strings rather than integers in most payment gateways (Stripe, PayPal, etc.). This change aligns the membership-create schema with the membership-update schema which was already corrected.

###   Changes:
  - inc/api/schemas/membership-create.php: Change gateway_customer_id type from 'integer' to 'string'

  This resolves API validation errors when integrating with external tools like N8N for membership creation workflows.


superdav42 (approved):
I thought I already merged this is, but I guess not.

Issue #126, opened 2025-08-03
[bug]  Email template and email template preview -> Broken custom logo


pgroot91:
- Image is not shown in the preview of the email template.
- Image is not shown when you are editing the email template. The ID of the image exists and is saved in the object.

**How to reproduce**

1. Navigate to your network administration area.
2. In the submenu of the Multisite WaaS, click on broadcasts link.
3. Click "System Emails" button.
4. Click "Email Template" button.
5. In the customizer you can select the setting to upload a custom logo. Toggle the setting and upload a custom logo.
6. Save the settings. 

The data is saved in the `wp_sitemeta` table and under the `wp-ultimo_email_template` meta_key.

```
[
  "use_custom_logo" => true,
  "custom_logo" => "http://8",
  "display_company_address" => true,
  "background_color" => "#f1f1f1",
  "title_color" => "#000000",
  "title_size" => "h3",
  "title_align" => "center",
  "title_font" => "Helvetica Neue, Helvetica, Helvetica, Arial, sans-serif",
  "content_color" => "#000000",
  "content_align" => "left",
  "content_font" => "Helvetica Neue, Helvetica, Helvetica, Arial, sans-serif",
  "footer_font" => "Helvetica Neue, Helvetica, Helvetica, Arial, sans-serif",
  "footer_text" => "",
  "footer_color" => "#000000",
  "footer_align" => "center",
]
```

The value being saved is not the URL but the image ID. It is a mess in the following files

```
inc\admin-pages\class-email-template-customize-admin-page.php
inc\admin-pages\class-template-previewer-customize-admin-page.php
inc\admin-pages\class-invoice-template-customize-admin-page.php
```

You see that the logo is broken and a URL is missing in the template preview.

<img width="3790" height="1977" alt="Image" src="https://github.com/user-attachments/assets/3e95cc63-809f-42a0-a76d-eeca095f91af" />

pgroot91:
Also broken in the invoice template.

PR #153, opened 2025-08-07
fix email templates and invoice template custom logo

pgroot91:
Solves #126 

I'm still not able to test emails locally so not sure if it is 100% working but the issues in the settings and UI seems to be solved.


pgroot91:
At the moment investigation the invoice. Don't have a solution yet.

PR #159, opened 2025-08-09
Fix: Add defensive checks for Slim SEO plugin compatibility

vuckro:
- Add null checks in UI element output methods
- Prevent errors when site/membership objects are null
- Ensures compatibility with Slim SEO plugin that calls output() directly
- Defensive programming approach for robustness

### DIFF
--- a/inc/ui/class-account-summary-element.php
+++ b/inc/ui/class-account-summary-element.php
@@ -294,6 +294,11 @@ class Account_Summary_Element extends Base_Element {
 	 */
 	public function output($atts, $content = null) {
 
+		// Defensive check - setup() may have been called but site can still be null
+		if ( ! $this->site) {
+			return '';
+		}
+
 		$atts = array_merge($atts, $this->atts);
 
 		$atts['site'] = $this->site;
--- a/inc/ui/class-account-summary-element.php
+++ b/inc/ui/class-account-summary-element.php
@@ -296,6 +296,7 @@ class Account_Summary_Element extends Base_Element {
 
 		// Defensive check - setup() may have been called but site can still be null
 		if ( ! $this->site) {
+			_doing_it_wrong(__METHOD__, esc_html__('setup() or setup_preview() must be called before output().', 'multisite-ultimate'));
 			return '';
 		}
 
--- a/inc/ui/class-billing-info-element.php
+++ b/inc/ui/class-billing-info-element.php
@@ -272,6 +272,11 @@ class Billing_Info_Element extends Base_Element {
 	 */
 	public function output($atts, $content = null) {
 
+		// Defensive check - setup() may have been called but membership can still be null
+		if ( ! $this->membership) {
+			return '';
+		}
+
 		$atts['membership'] = $this->membership;
 
 		$atts['billing_address'] = $this->membership->get_billing_address();
--- a/inc/ui/class-billing-info-element.php
+++ b/inc/ui/class-billing-info-element.php
@@ -274,6 +274,7 @@ class Billing_Info_Element extends Base_Element {
 
 		// Defensive check - setup() may have been called but membership can still be null
 		if ( ! $this->membership) {
+			_doing_it_wrong(__METHOD__, esc_html__('setup() or setup_preview() must be called before output().', 'multisite-ultimate'));
 			return '';
 		}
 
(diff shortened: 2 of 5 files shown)


vuckro:
I'm trying to update with the recommendations quickly.

superdav42:
Now that I see the problem I think you just need to call this function `[wu_element_setup_preview](https://github.com/superdav42/wp-multisite-waas/blob/main/inc/functions/element.php#L18)()` in the appropriate action. It seems to be an issue when bricks and slim seo is active. Probably you just need to to setup the preview in an action fired by the bricks plugin when it loads the editor. 

See https://github.com/superdav42/wp-multisite-waas/blob/main/inc/compat/class-elementor-compat.php for how a class is setup to fix a similar problem in elementor. Particularly this line:
```
		add_action('elementor/widget/shortcode/skins_init', [$this, 'maybe_setup_preview']);
```
If the preview is setup correctly none of the other code changes in the other files will be necessary and they will show appropriate preview data. You should make a single class like Bricks_Compat which will keep all the code in once place.

vuckro:
Thanks for the feedback and explanations.

I don't think Bricks is the problem (same error with only Slim SEO active + Multisite Ultimate).

Also, the approach with a compatibility class causes problems in my tests.
I'm going to revert to the original defensive changes, which were safer.

Hoping to hear back on [this slim seo issue](https://github.com/elightup/slim-seo/issues/232).

inc/ui/class-account-summary-element.php
```diff
@@ -294,6 +294,11 @@ public function setup_preview(): void {
 	 */
 	public function output($atts, $content = null) {
 
+		// Defensive check - setup() may have been called but site can still be null
+		if ( ! $this->site) {
+			return '';
```
superdav42:
```suggestion
		if ( ! $this->site) {
			_doing_it_wrong(__METHOD__, esc_html__('setup() or setup_preview() must be called before output().', 'multisite-ultimate'));
			return '';
```

inc/ui/class-billing-info-element.php
```diff
@@ -272,6 +272,11 @@ public function setup_preview(): void {
 	 */
 	public function output($atts, $content = null) {
 
+		// Defensive check - setup() may have been called but membership can still be null
+		if ( ! $this->membership) {
+			return '';
```
superdav42, line 278:
```suggestion
			_doing_it_wrong(__METHOD__, esc_html__('setup() or setup_preview() must be called before output().', 'multisite-ultimate'));
			return '';
```

inc/ui/class-current-site-element.php
```diff
@@ -352,6 +352,11 @@ public function register_scripts(): void {
 	 */
 	public function output($atts, $content = null) {
 
+		// Defensive check - setup() may have been called but site can still be null
+		if ( ! $this->site) {
+			return '';
```
superdav42, line 358:
```suggestion
			_doing_it_wrong(__METHOD__, esc_html__('setup() or setup_preview() must be called before output().', 'multisite-ultimate'));
			return '';
```

inc/ui/class-invoices-element.php
```diff
@@ -274,6 +274,11 @@ public function setup_preview(): void {
 	 */
 	public function output($atts, $content = null) {
 
+		// Defensive check - setup() may have been called but membership can still be null
+		if ( ! $this->membership) {
+			return '';
```
superdav42, line 280:
```suggestion
			_doing_it_wrong(__METHOD__, esc_html__('setup() or setup_preview() must be called before output().', 'multisite-ultimate'));
			return '';
```

inc/ui/class-limits-element.php
```diff
@@ -240,6 +240,11 @@ public function setup_preview(): void {
 	 */
 	public function output($atts, $content = null) {
 
+		// Defensive check - setup() may have been called but site can still be null
+		if ( ! $this->site) {
+			return '';
```
superdav42, line 246:
```suggestion
			_doing_it_wrong(__METHOD__, esc_html__('setup() or setup_preview() must be called before output().', 'multisite-ultimate'));
			return '';
```

vuckro:
@superdav42 It seems that the merge did not completely resolve the issue with my first code proposal:



Can you take a look? I can create a temporary site if necessary.

Also following [this issue #232](https://github.com/elightup/slim-seo/issues/232), but no response for now

PR #160, opened 2025-08-09
Fix : Add customer meta deletion functionality with simple delete link

vuckro:
Adds ability to delete orphaned customer custom meta fields from the Customer Edit screen.
Includes nonce verification, safe redirects, and inline "Supprimer" link only for fields without a form reference.


vuckro:
@superdav42 The problem seems to persist in 2.4.3.

PR #161, opened 2025-08-09
Fix : API post payment

vuckro:
Corrects product publication to a payment with the API. This prevents fatal errors when creating payments via the API, as $this->line_items always contains valid objects instead of arrays or null values.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed inconsistent handling of checkout line items when loading payments, ensuring items are properly recognized and displayed.
  * Resolved cases where totals, taxes, and discounts could appear incorrect due to misloaded items, improving accuracy across checkout, invoices, receipts, and emails.
  * Improved stability for payment edits, refunds, and reports by standardizing line item processing.


superdav42:
I already committed a fix which converts arrays to a class when calling set_line_items() here https://github.com/superdav42/wp-multisite-waas/commit/14931bf2158c9c7f9b559cf33a62829ceb8b008d#diff-0e9e32e96750797b179eda709b0010fc22de54fc734fbbd82c13032b9d5e974bR621
It should have already been in the last release. Can you check if this fixed your problems?

It won't be retroactive so if there is already bad data in a membership it will need to be cleaned somehow. But this should prevent any bad data from being stored in the database so it can be unserialized correctly

PR #269, opened 2025-11-15
Improve post type registration logic

tamkendigital:
Fix undefined property warnings and improve data validation in  register_emulated_post_types() method.

This commit addresses several issues:

1. **Fixed undefined property warnings**: Previously, the code  attempted to access $pt->post_type and $pt->label properties  without verifying that the data structure was valid, causing  "Undefined property: stdClass::$post_type" warnings when  corrupted data existed in the database.

2. **Added automatic data cleanup**: The method now automatically  cleans up corrupted entries (non-array items or entries missing  required keys) and saves the cleaned data back to the database.

3. **Improved validation**: Added proper validation to ensure that:
   - Each entry is a valid array
   - Required keys (post_type and label) exist
   - Empty values are allowed for new entries (to support the UI  workflow where users add empty entries first, then fill them in)
   - Only entries with non-empty values are registered as post types

4. **Enhanced security**: Added sanitization using sanitize_key()  and sanitize_text_field() for all user input before processing  and registration.

5. **Better error handling**: The method now gracefully handles  corrupted data by skipping invalid entries instead of throwing  errors, allowing users to delete post types without issues.

The fix ensures that:
- Users can add new post types without errors
- Users can delete post types without warnings
- Corrupted data is automatically cleaned up
- The registration process only attempts to register valid,  non-empty post types

Fixes issues where users were unable to delete post types after  adding them, and prevents undefined property warnings in the  debug log.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved post-type data handling: validates and sanitizes saved entries, skips invalid or empty definitions, and exits early when none are present.
  * Automatically cleans and persists corrected settings when corrupted or incomplete entries are found.
  * Continues to preserve and skip already-registered valid post types to maintain consistent behavior.

### DIFF
--- a/inc/limits/class-post-type-limits.php
+++ b/inc/limits/class-post-type-limits.php
@@ -88,7 +88,7 @@ class Post_Type_Limits {
 
 		foreach ($emulated_post_types as $index => $pt) {
 			// Verify that $pt is an array
-			if (!is_array($pt)) {
+			if ( ! is_array($pt)) {
 				$needs_update = true;
 				continue;
 			}


inc/limits/class-post-type-limits.php
```diff
@@ -78,29 +78,75 @@ public function register_emulated_post_types(): void {
...
+
+		foreach ($emulated_post_types as $index => $pt) {
+			// Verify that $pt is an array
+			if (!is_array($pt)) {
+				$needs_update = true;
+				continue;
+			}
+
+			// Verify that required keys exist (allow empty values for new entries)
+			if (!isset($pt['post_type']) || !isset($pt['label'])) {
```
superdav42:
```suggestion
			if ( ! isset($pt['post_type']) || ! isset($pt['label'])) {
```

inc/limits/class-post-type-limits.php
```diff
@@ -78,29 +78,75 @@ public function register_emulated_post_types(): void {
...
+			return;
+		}
+
+		// Clean up corrupted data automatically
+		$cleaned_post_types = [];
+		$needs_update       = false;
+
+		foreach ($emulated_post_types as $index => $pt) {
+			// Verify that $pt is an array
+			if (!is_array($pt)) {
```
superdav42:
```suggestion
			if ( ! is_array($pt)) {
```

superdav42:
@tamkendigital Thanks for the PR! It works perfectly. I just made some code style changes and merged it. It'll be in the next release.

PR #278, opened 2025-12-01
fix(domain): prevent multiple primary domains per site

mahyarrezghi:
- Track original primary state in Domain model to detect changes
- Unset other primary domains when setting a new primary
- Ensure consistency across API, admin, and creation flows

## Summary by CodeRabbit

* **New Features**
  * Automatic cleanup of previous primary domain assignments when a new primary is set.

* **Behavior Changes**
  * Improved detection of when a domain becomes primary to ensure updates run only when needed.
  * Cleanup now consistently runs as part of the domain save process (removed redundant immediate triggers elsewhere), reducing duplicate background tasks.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/inc/admin-pages/class-domain-list-admin-page.php
+++ b/inc/admin-pages/class-domain-list-admin-page.php
@@ -241,22 +241,6 @@ class Domain_List_Admin_Page extends List_Admin_Page {
 				wp_send_json_error($domain);
 			}
 
-			if (wu_request('primary_domain')) {
-				$old_primary_domains = wu_get_domains(
-					[
-						'primary_domain' => true,
-						'blog_id'        => wu_request('blog_id'),
-						'id__not_in'     => [$domain->get_id()],
-						'fields'         => 'ids',
-					]
-				);
-
-				/*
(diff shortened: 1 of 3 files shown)


nourihsc:
Hello, and thank you for developing this plugin. I found it while searching the plugin repository and installed it to test it out, but I encountered the following message during the second step: **"Create Database Tables – Ultimate Multisite uses custom tables for performance reasons. We need to create those tables and make sure they are setup properly before we can activate the plugin. Installation of the table failed."** Could you please provide some guidance on how to proceed with the installation and explain the cause of this error?

nourihsc:
> Hello, and thank you for developing this plugin. I found it while searching the plugin repository and installed it to test it out, but I encountered the following message during the second step: **"Create Database Tables – Ultimate Multisite uses custom tables for performance reasons. We need to create those tables and make sure they are setup properly before we can activate the plugin. Installation of the table failed."** Could you please provide some guidance on how to proceed with the installation and explain the cause of this error?

The issue was resolved by changing the storage engine of the tables to InnoDB.

PR #792, opened 2026-04-11
fix: auto-reset stale is_publishing flag to prevent infinite Creating spinner

superdav42:
## Summary

Fixes the infinite "Creating" spinner on the thank-you page when the site creation process is killed mid-execution.

**Root cause:** `publish_pending_site()` sets `is_publishing=true` on the pending site object and saves it to membership meta. If the PHP process dies (OOM, timeout, server restart) after the flag is set but before the site creation completes or error handlers run, the flag stays `true` permanently. The AJAX poll handler (`wu_check_pending_site_created`) returns `running` forever, and the Action Scheduler retry bails out because it sees the flag and assumes another process is already handling it.

**Fix:**
- Add `publishing_started_at` timestamp to the Site model, automatically recorded when `set_publishing(true)` is called
- Add `is_publishing_stale($timeout)` method — returns true when publishing has been running longer than the timeout (default 5 minutes)
- `check_pending_site_created()` now detects stale state, resets the flag, re-enqueues the async action, and returns `stopped` so the JS can retry
- Pre-2.5.3 serialized pending site objects (which lack the timestamp field) are treated as stale when `is_publishing` is true, since the process that set it is clearly gone

**Files changed:**
- `inc/models/class-site.php` — new property, getter, and staleness check method
- `inc/managers/class-membership-manager.php` — staleness detection in AJAX handler
- `tests/WP_Ultimo/Models/Site_Test.php` — 7 new test cases for the staleness logic

**Testing instructions:**
1. Install the build from this branch
2. Go through a WooCommerce checkout flow that creates a site
3. Verify the thank-you page transitions from "Creating" to the completed state
4. To test the stuck state recovery: if a site is currently stuck in "Creating", this fix should auto-recover within 5 minutes of installing

## Release Notes

* **New Features**
  * Implemented automatic detection and recovery for site publishing processes that become stalled or unresponsive
  * Publishing tasks now automatically resume when a stalled state is detected beyond the timeout threshold

* **Bug Fixes**
  * Resolved issues where sites could remain indefinitely stuck in an incomplete publishing state

* **Tests**
  * Added comprehensive unit tests validating publishing state detection, timeout handling, and automatic recovery mechanisms

### DIFF
--- a/inc/managers/class-membership-manager.php
+++ b/inc/managers/class-membership-manager.php
@@ -167,6 +167,11 @@ class Membership_Manager extends Base_Manager {
 	/**
 	 * Processes a delayed site publish action.
 	 *
+	 * Checks whether the pending site has been created. If is_publishing
+	 * has been true for longer than 5 minutes, we assume the creation
+	 * process was killed (PHP timeout, OOM, server restart) and reset
+	 * the flag so the Action Scheduler retry can pick it up.
+	 *
 	 * @since 2.0.11
 	 */
 	public function check_pending_site_created() {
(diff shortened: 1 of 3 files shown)


superdav42:
## Summary
Fixes the infinite "Creating" spinner on the thank-you page when the site creation process is killed mid-execution.
**Root cause:** `publish_pending_site()` sets `is_publishing=true` on the pending site object and saves it to membership meta. If the PHP process dies (OOM, timeout, server restart) after the flag is set but before the site creation completes or error handlers run, the flag stays `true` permanently. The AJAX poll handler (`wu_check_pending_site_created`) returns `running` forever, and the Action Scheduler retry bails out because it sees the flag and assumes another process is already handling it.
**Fix:**
- Add `publishing_started_at` timestamp to the Site model, automatically recorded when `set_publishing(true)` is called
- Add `is_publishing_stale($timeout)` method — returns true when publishing has been running longer than the timeout (default 5 minutes)
- `check_pending_site_created()` now detects stale state, resets the flag, re-enqueues the async action, and returns `stopped` so the JS can retry
- Pre-2.5.3 serialized pending site objects (which lack the timestamp field) are treated as stale when `is_publishing` is true, since the process that set it is clearly gone
**Files changed:**
- `inc/models/class-site.php` — new property, getter, and staleness check method
- `inc/managers/class-membership-manager.php` — staleness detection in AJAX handler
- `tests/WP_Ultimo/Models/Site_Test.php` — 7 new test cases for the staleness logic
**Testing instructions:**
1. Install the build from this branch
2. Go through a WooCommerce checkout flow that creates a site
3. Verify the thank-you page transitions from "Creating" to the completed state
4. To test the stuck state recovery: if a site is currently stuck in "Creating", this fix should auto-recover within 5 minutes of installing

---
Merged via PR #792 to main.

---
[aidevops.sh](https://aidevops.sh) v3.6.235 spent 7m on this as a headless bash routine.

PR #843, opened 2026-04-14
Change 'checked' to 'selected' for select list options

commercial-hippie:
Pre-selected values don't show correctly in the checkout forms, currently printing out "checked" instead of "selected".

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed select dropdown fields in checkout to properly display selected options during form submission.

### DIFF
--- a/views/checkout/fields/field-select.php
+++ b/views/checkout/fields/field-select.php
@@ -42,7 +42,7 @@ defined('ABSPATH') || exit;
 
 	<?php if ($field->placeholder) : ?>
 
-	<option <?php checked(! $field->value); ?> class="wu-opacity-75"><?php echo esc_html($field->placeholder); ?></option>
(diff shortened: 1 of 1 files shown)


superdav42:
@commercial-hippie Thank you so much for this catch — this is a fantastic contribution! This bug has been lurking in the codebase since the very first commit (November 2024) and survived through seven subsequent commits over 16 months without anyone noticing. Using `checked()` instead of `selected()` on `<option>` elements meant the select fields were never properly reflecting their selected state in the HTML output.

Really appreciate you taking the time to find this, fix it, and submit the PR. Merged!

We'll follow up with a small additional improvement based on CodeRabbit's review (adding an explicit `value=""` on the placeholder option and tightening the falsy comparison), but your fix is the one that matters.

Thanks again for contributing to Ultimate Multisite!

PR #852, opened 2026-04-14
GH#837: fix network activation reliability in multisite setup wizard

### LINKED ISSUE #837: The multi-site settup wizard does not network activate the plugin
Image
The last step of this page should network activate the ultimate multisite plugin but it does not. It appears to but it really doesn't. sometimes.

superdav42:
## Summary

Fixes a race condition in the Multisite Setup Wizard where the final **Network Activate Plugin** step would sometimes silently fall back to single-site activation instead of network-wide activation.

### Root cause

`_install_network_activate()` in `Multisite_Network_Installer` called `activate_plugin(WP_ULTIMO_PLUGIN_FILE, '', true)` to perform network activation. WordPress's `activate_plugin()` checks `is_multisite()` internally and uses the `active_sitewide_plugins` sitemeta key **only when multisite is active**. If `is_multisite()` returns `false`, it falls back silently to single-site activation.

The MULTISITE constant is written to `wp-config.php` in the immediately-preceding step (`_install_update_wp_config`). Each wizard step runs in a separate AJAX request, but if OPcache or another PHP bytecode cache serves a stale `wp-config.php` in the next request, `is_multisite()` returns `false` in the `network_activate` step — producing the "appears to but doesn't, sometimes" behaviour from the issue.

A secondary bug: the early-return guard used `is_plugin_active(WP_ULTIMO_PLUGIN_FILE)` with the full absolute path instead of the plugin basename (`WP_ULTIMO_PLUGIN_BASENAME`), so the guard never fired.

### Fix

Replace `activate_plugin()` with a direct write to the `wp_sitemeta` table, as the existing code comment already described as the intent. This bypasses the `is_multisite()` dependency entirely and guarantees reliable network activation regardless of bytecode caching.

## Files changed

- **EDIT: `inc/installers/class-multisite-network-installer.php`** — rewrites `_install_network_activate()` to write directly to `{prefix}sitemeta` via `$wpdb->get_row` / `$wpdb->update` / `$wpdb->insert`, using `WP_ULTIMO_PLUGIN_BASENAME` throughout
- **NEW: `tests/WP_Ultimo/Installers/Multisite_Network_Installer_Test.php`** — 16 tests covering `get_steps()`, `get_config()`, `check_network_tables_exist()`, and all branches of `_install_network_activate()` including idempotency and existing-plugin preservation

## Verification

```bash
WP_TESTS_DIR=/tmp/wordpress-tests-lib vendor/bin/phpunit \
  --filter Multisite_Network_Installer_Test --no-coverage --testdox
# OK (16 tests, 49 assertions)

vendor/bin/phpstan analyse inc/installers/class-multisite-network-installer.php --no-progress
# [OK] No errors
```

Resolves #837

### DIFF
--- a/inc/installers/class-multisite-network-installer.php
+++ b/inc/installers/class-multisite-network-installer.php
@@ -256,27 +256,73 @@ class Multisite_Network_Installer extends Base_Installer {
 	}
 
 	/**
-	 * Step 5: Network-activate Ultimate Multisite.
+	 * Step 4: Network-activate Ultimate Multisite.
 	 *
-	 * Writes directly to the sitemeta table because multisite
-	 * is not yet active in the current PHP process.
+	 * Writes directly to the sitemeta table because the MULTISITE constant
+	 * was written to wp-config.php only in the previous step
+	 * (_install_update_wp_config) and may not yet be reflected in the current
+	 * PHP process when OPcache or another bytecode cache is active. Using
+	 * activate_plugin() would silently fall back to single-site activation
+	 * when is_multisite() returns false. Bypassing it and writing directly to
+	 * sitemeta guarantees network-wide activation regardless of whether
+	 * multisite constants are loaded in this process.
 	 *
 	 * @since 2.0.0
-	 * @throws \Exception When the activation fails.
+	 * @throws \Exception When the sitemeta write fails.
 	 * @return void
 	 */
 	public function _install_network_activate(): void { // phpcs:ignore PSR2.Methods.MethodDeclaration.Underscore
-		// If already active, succeed early.
-		if (is_plugin_active(WP_ULTIMO_PLUGIN_FILE)) {
+
+		global $wpdb;
+
+		$sitemeta_table = $wpdb->base_prefix . 'sitemeta';
+
+		// phpcs:disable WordPress.DB.DirectDatabaseQuery.DirectQuery, WordPress.DB.DirectDatabaseQuery.NoCaching
+		$existing = $wpdb->get_row(
+			$wpdb->prepare(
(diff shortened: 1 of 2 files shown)


superdav42:
## Merge Summary

**Issue:** #837 — The multisite setup wizard does not network activate the plugin

**Root cause:** `_install_network_activate()` in `Multisite_Network_Installer` called `activate_plugin()`, which internally checks `is_multisite()`. The MULTISITE constant is written to `wp-config.php` only in the preceding step; if OPcache serves a stale copy on the next AJAX request, `is_multisite()` returns `false` and the plugin is activated single-site only — explaining the intermittent failure.

A secondary bug: the early-return guard used `WP_ULTIMO_PLUGIN_FILE` (full absolute path) instead of `WP_ULTIMO_PLUGIN_BASENAME`, so the guard never fired.

**Fix:** Replaced `activate_plugin()` with a direct write to the `{prefix}sitemeta` table, bypassing the `is_multisite()` dependency entirely. The implementation matches what the original code comment already described as the intent.

**Changes:**
- `inc/installers/class-multisite-network-installer.php` — rewrote `_install_network_activate()`
- `tests/WP_Ultimo/Installers/Multisite_Network_Installer_Test.php` — 16 new tests (49 assertions)

**Verification:** All 16 new tests pass; PHPStan reports no errors.



---
[aidevops.sh](https://aidevops.sh) v3.8.25 plugin for [OpenCode](https://opencode.ai) v1.4.3 with claude-sonnet-4-6 spent 12m and 1,346 tokens on this as a headless worker. Overall, 1m since this issue was created.

PR #866, opened 2026-04-14
fix: default role in Login & Registration settings shows administrator by default

### LINKED ISSUE #865: The default role in settings should be administrator
Here
Image
In the login and registrations settings the default role should be defaulted to administrator. The "Ultimate Multisite Default" option should not be an option. In the initial state it should be Administrator. I believe this changed when we made some optimizations to how settings are loaded. We need each get setting option function call to use the correct default value in the default parameter

superdav42:
## Problem

On a fresh install (no settings saved to DB), the **Default Role** select in the Login & Registration settings page showed **"Use Ultimate Multisite default"** as the visible selection instead of **"Administrator"**.

Two bugs combined to cause this:

### Bug 1 — Unwanted "Ultimate Multisite default" option injected

`Field::__get()` calls `wu_get_roles_as_options` via `call_user_func($attr, $this)`, forwarding the `Field` object as the first argument. Since an object is truthy, `wu_get_roles_as_options($field_object)` resolved `$add_default_option = true`, injecting the placeholder "Use Ultimate Multisite default" option (key `'default'`).

**Fix:** replace the bare string reference with an explicit closure:
```php
// Before
'options' => 'wu_get_roles_as_options',

// After
'options' => fn() => wu_get_roles_as_options(),
```
Applied to both `default_role` and `main_site_default_role` fields.

### Bug 2 — Vue data-state missing `default_role` on fresh install

`get_all_with_defaults()` guarded its update loop with `isset($all_settings[$field_slug])`. When the DB has no saved settings the key is absent, so the Vue `data-state` JSON never received the `default_role` entry. Vue's v-model had no value to bind and fell back to the first rendered option (the injected placeholder from Bug 1).

**Fix:** drop the `isset()` guard so registered fields are always included in the state with their computed default value:
```php
// Before
if (isset($all_settings[$field_slug]) && $value !== $all_settings[$field_slug]) {

// After
if (!isset($all_settings[$field_slug]) || $value !== $all_settings[$field_slug]) {
```

## Files changed

- `EDIT: inc/class-settings.php` — closures for both role select `options`, updated `get_all_with_defaults` condition
- `EDIT: tests/WP_Ultimo/Settings_Test.php` — 5 new assertions covering both regressions

## Verification

```bash
WP_TESTS_DIR=/tmp/wordpress-tests-lib vendor/bin/phpunit --filter Settings_Test
# OK (47 tests, 63 assertions)
```

Resolves #865

### DIFF
--- a/inc/class-settings.php
+++ b/inc/class-settings.php
@@ -176,7 +176,7 @@ class Settings implements \WP_Ultimo\Interfaces\Singleton {
 			foreach ($section['fields'] ?? [] as $field_slug => $field_atts) {
 				if (is_callable($field_atts['value'])) {
 					$value = $field_atts['value']();
-					if (isset($all_settings[ $field_slug ]) && $value !== $all_settings[ $field_slug ]) {
+					if ( ! isset($all_settings[ $field_slug ]) || $value !== $all_settings[ $field_slug ]) {
 						$all_settings[ $field_slug ] = $value;
 					}
 				}
@@ -934,7 +934,7 @@ class Settings implements \WP_Ultimo\Interfaces\Singleton {
 				'desc'    => __('Set the role to be applied to the user during the signup process.', 'ultimate-multisite'),
 				'type'    => 'select',
 				'default' => 'administrator',
-				'options' => 'wu_get_roles_as_options',
+				'options' => fn() => wu_get_roles_as_options(),
 			]
 		);
 
@@ -957,7 +957,7 @@ class Settings implements \WP_Ultimo\Interfaces\Singleton {
 				'desc'    => __('Select the role Ultimate Multisite should use when adding the user to the main site of your network. Be careful.', 'ultimate-multisite'),
 				'type'    => 'select',
 				'default' => 'subscriber',
-				'options' => 'wu_get_roles_as_options',
+				'options' => fn() => wu_get_roles_as_options(),
 				'require' => [
 					'add_users_to_main_site' => 1,
 				],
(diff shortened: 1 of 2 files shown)


superdav42:
## Merge Summary

**Issue:** #865 — Default role in Login & Registration settings should be Administrator

**Root cause (two bugs):**

1. `Field::__get()` called `wu_get_roles_as_options` via `call_user_func($attr, $this)`, passing the `Field` object as `$add_default_option`. Since an object is truthy, the "Use Ultimate Multisite default" placeholder option was always injected into the role select dropdowns.

2. `get_all_with_defaults()` had an `isset()` guard that prevented settings not yet in the DB from being included in the Vue `data-state` JSON. On a fresh install, `default_role` was absent from state, leaving the dropdown with no bound value — it displayed the first option (the injected placeholder from bug 1).

**Changes:**
- `inc/class-settings.php`: Changed `'options' => 'wu_get_roles_as_options'` to `'options' => fn() => wu_get_roles_as_options()` for both `default_role` and `main_site_default_role` fields. Changed `get_all_with_defaults` condition from `isset(...) &&` to `!isset(...) ||` to ensure defaults are included for unsaved settings.
- `tests/WP_Ultimo/Settings_Test.php`: Added 5 new tests covering both regressions.

**Verified:** `WP_TESTS_DIR=/tmp/wordpress-tests-lib vendor/bin/phpunit --filter Settings_Test` → OK (47 tests, 63 assertions)

PR #1028, opened 2026-04-30
GH#1009: Fix import cron schedule circular dependency

### LINKED ISSUE #1009: Import cron schedule has circular dependency — async imports may silently never execute
## Bug

The cron schedule registration for imports has a circular dependency that prevents the schedule from ever being created when there ARE pending imports.

In `maybe_add_schedule()` (line 1467-1481):

```php
public function maybe_add_schedule(array $schedules): array {
    $pending_imports = wu_exporter_get_pending_imports();
    if (empty($pending_imports)) {
        return $schedules;  // <-- returns early when no imports
    }
    $schedules['wu_site_every_minute'] = [...];
    return $schedules;
}
```

And in `maybe_run_imports()` (line 1452-1457):

```php
public function maybe_run_imports(): void {
    if (! wp_next_scheduled('wu_import_site')) {
        wp_schedule_event(time() + 10, 'wu_site_every_minute', 'wu_import_site');
    }
}
```

The problem:

1. `maybe_run_imports()` runs on `init` and tries to schedule the event with the `wu_site_every_minute` interval
2. But `maybe_add_schedule()` runs on `cron_schedules` which fires DURING `wp_schedule_event()` 
3. If there are no pending imports when the schedule is first checked, the custom schedule interval is never registered
4. When an import is queued, `wp_schedule_event()` may fail because `wu_site_every_minute` hasn't been registered yet (the schedule filter runs during `wp_schedule_event`, so timing is fragile)

Additionally, the import handler `handle_site_import()` is hooked to the `wu_import_site` action but with wrong parameters — it accepts no params while the hook could fire with params depending on how WP cron calls it.

## Impact

Imports queued via the admin UI may never actually execute because the cron job setup is unreliable.

## Suggested Fix

1. Always register the `wu_site_every_minute` schedule (remove the pending imports check in `maybe_add_schedule`)
2. Or use WordPress's built-in `wp_schedule_single_event()` when an import is queued, instead of relying on a recurring schedule

### Files to modify

- `EDIT: inc/site-exporter/class-site-exporter.php:1467-1481` — always register the cron schedule
- `EDIT: inc/site-exporter/class-site-exporter.php:1452-1457` — consider using single events per import

## Severity

**High** — async imports may silently never execute.

---
[aidevops.sh](https://aidevops.sh) v3.13.11 plugin for [OpenCode](https://opencode.ai) v1.3.17 with claude-opus-4-6 spent 6m and 14,562 tokens on this as a headless worker.

superdav42:
## Summary

Fixes the circular dependency in the import cron schedule registration that caused async imports to silently never execute.

## Root Cause

`maybe_add_schedule()` (hooked to `cron_schedules` filter) returned early when no pending imports existed. Since WordPress evaluates the `cron_schedules` filter *inside* `wp_schedule_event()`, this meant:

1. On a fresh site (no imports queued), `maybe_run_imports()` runs on `init` and calls `wp_schedule_event()`.
2. Inside `wp_schedule_event()`, the `cron_schedules` filter fires.
3. `maybe_add_schedule()` finds no pending imports → returns early → `wu_site_every_minute` is never registered.
4. `wp_schedule_event()` cannot validate the schedule name and silently fails.
5. No WP cron event is ever scheduled for imports.

When an import is finally queued, the cron event needed to process it was never registered.

## Changes

**`inc/site-exporter/class-site-exporter.php`**

- **`maybe_add_schedule()`**: Remove the pending-imports guard — always register the `wu_site_every_minute` interval. The `handle_site_import()` callback already returns `false` early when there are no pending imports, so an idle recurring event is safe.

- **`setup()`**: Use `add_filter()` for `cron_schedules` (it is a filter, not an action).

- **`setup()`**: Remove the incorrect `$accepted_args=3` from the `wu_import_site` → `handle_site_import` registration. `handle_site_import()` accepts no parameters.

## Verification

- PHP syntax: `php -l inc/site-exporter/class-site-exporter.php` — clean
- The `wu_site_every_minute` schedule is now always available, so `wp_schedule_event()` in `maybe_run_imports()` will succeed on every request.

Resolves #1009

---
[aidevops.sh](https://aidevops.sh) v3.13.12 plugin for [OpenCode](https://opencode.ai) v1.3.17 with claude-sonnet-4-6 spent 7m and 11,567 tokens on this as a headless worker.

### DIFF
--- a/inc/site-exporter/class-site-exporter.php
+++ b/inc/site-exporter/class-site-exporter.php
@@ -70,11 +70,11 @@ final class Site_Exporter {
 
 		add_action('wu_export_site', [$this, 'handle_site_export'], 10, 3);
 
-		add_action('wu_import_site', [$this, 'handle_site_import'], 10, 3);
+		add_action('wu_import_site', [$this, 'handle_site_import']);
 
 		add_filter('wu_site_exporter_files_to_zip', [$this, 'maybe_exclude_wp_ultimo_plugins']);
 
-		add_action('cron_schedules', [$this, 'maybe_add_schedule']);
+		add_filter('cron_schedules', [$this, 'maybe_add_schedule']);
 
 		add_action('init', [$this, 'maybe_run_imports']);
 
@@ -1457,7 +1457,16 @@ final class Site_Exporter {
 	}
 
 	/**
-	 * Maybe adds a new schedule.
+	 * Adds the custom cron schedule interval.
+	 *
+	 * Always registers the wu_site_every_minute interval so that
+	 * wp_schedule_event() can succeed regardless of whether there are
+	 * pending imports at the time the cron_schedules filter runs.
+	 * Previously this method returned early when no imports were pending,
+	 * creating a circular dependency: the schedule was only registered when
+	 * imports existed, but the event could not be scheduled until the
+	 * interval was registered — meaning the very first import might never
+	 * be processed.
 	 *
 	 * @since 2.5.0
 	 *
@@ -1465,12 +1474,6 @@ final class Site_Exporter {
 	 * @return array
 	 */
 	public function maybe_add_schedule(array $schedules): array {
-
-		$pending_imports = wu_exporter_get_pending_imports();
-
-		if (empty($pending_imports)) {
-			return $schedules;
-		}
 
 		$schedules['wu_site_every_minute'] = [
 			'interval' => 60,


superdav42:
## Merge Summary

**Issue:** #1009 — Import cron schedule has circular dependency
**Branch:** `feature/auto-20260430-002623-gh1009`
**Commit:** 7e54ce2e

### What was done

Removed the circular dependency in the import cron schedule registration in `inc/site-exporter/class-site-exporter.php`:

1. **`maybe_add_schedule()`**: Removed the early return when no pending imports exist. The `wu_site_every_minute` cron interval is now always registered. Previously, this guard meant that on a fresh site, `wp_schedule_event()` would fail silently because the custom interval was missing when the `cron_schedules` filter fired internally during the scheduling call.

2. **`setup()`**: Changed `add_action('cron_schedules', ...)` → `add_filter('cron_schedules', ...)` since `cron_schedules` is a filter hook.

3. **`setup()`**: Removed incorrect `$accepted_args=3` from the `wu_import_site` → `handle_site_import` hook registration; the handler accepts no parameters.

### How verified

- `php -l inc/site-exporter/class-site-exporter.php` — no syntax errors
- `handle_site_import()` returns false early when no pending imports, so the idle recurring cron event is safe



---
[aidevops.sh](https://aidevops.sh) v3.13.12 plugin for [OpenCode](https://opencode.ai) v1.3.17 with claude-sonnet-4-6 spent 8m and 13,041 tokens on this as a headless worker.

PR #1042, opened 2026-04-30
test: add regression tests for Site_Exporter cron circular dependency (GH#1009)

superdav42:
## Summary

Adds a `Site_Exporter_Test` covering the `maybe_add_schedule()` cron circular dependency that was fixed in GH#1009 / PR #1028. No regression tests shipped with the original fix — this PR closes that gap.

## What

- New test file `tests/WP_Ultimo/Site_Exporter_Test.php` (10 tests, 18 assertions)
- Tests verify `maybe_add_schedule()` always registers `wu_site_every_minute` regardless of pending imports state
- Core regression test: `wp_schedule_event()` with the plugin interval succeeds even when no imports are pending (the silent failure case from the original bug)
- Covers hook registration, schedule idempotency, and existing-schedule preservation

## Why

The original fix (PR #1028) removed the conditional check that caused `wp_schedule_event()` to fail silently, but shipped without tests. Without a regression test, a future refactor could reintroduce the early-return pattern and break async imports again with no CI signal.

## Testing

```bash
vendor/bin/phpunit --filter Site_Exporter_Test
# OK (10 tests, 18 assertions)
```

Ref #1009

---
[aidevops.sh](https://aidevops.sh) v3.13.12 plugin for [OpenCode](https://opencode.ai) v1.3.17 with claude-sonnet-4-6 spent 13m and 42,906 tokens on this as a headless worker.

### DIFF
--- a/tests/WP_Ultimo/Site_Exporter_Test.php
+++ b/tests/WP_Ultimo/Site_Exporter_Test.php
@@ -1,0 +1,232 @@
+<?php
+/**
+ * Tests for Site_Exporter cron scheduling.
+ *
+ * Regression tests for GH#1009 — Import cron schedule circular dependency.
+ * The circular dependency was caused by maybe_add_schedule() returning early
+ * when no pending imports existed, preventing wp_schedule_event() from
+ * registering the wu_import_site event with a valid interval.
+ *
+ * @package WP_Ultimo\Site_Exporter
+ * @subpackage Tests
+ * @since 2.5.0
+ */
+
+namespace WP_Ultimo\Site_Exporter;
+
+use WP_UnitTestCase;
+
+/**
+ * Test class for Site_Exporter cron scheduling.
+ */
+class Site_Exporter_Test extends WP_UnitTestCase {
+
+	/**
+	 * @var Site_Exporter
+	 */
+	private Site_Exporter $exporter;
+
+	/**
+	 * Set up test fixtures.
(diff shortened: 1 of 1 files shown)


superdav42:
## MERGE_SUMMARY

**Issue**: GH#1009 — Import cron schedule circular dependency was already fixed in PR #1028 (merged 2026-04-30). This PR adds the missing regression tests.

**Root cause of the original bug**: `maybe_add_schedule()` returned early when no imports were pending, so the `wu_site_every_minute` cron interval was never registered. When `wp_schedule_event()` was called (in `maybe_run_imports()`), it silently failed because the interval was not yet registered.

**Fix status**: The fix (removing the early return from `maybe_add_schedule()`) was already applied and is on `main` as of commit `d5a1d0da`.

**This PR**: Adds 10 regression tests in `tests/WP_Ultimo/Site_Exporter_Test.php`:
- Core regression: `wp_schedule_event()` with `wu_site_every_minute` succeeds even when no imports are pending
- `maybe_add_schedule()` always registers the interval (guards against re-introducing the early return)
- Hook registration, schedule idempotency, and schedule preservation tests

**Verification**: `vendor/bin/phpunit --filter Site_Exporter_Test` → OK (10 tests, 18 assertions)


---
[aidevops.sh](https://aidevops.sh) v3.13.12 with claude-sonnet-4-6

PR #1110, opened 2026-05-05
feat(template-switch): redesign customer panel with current-template card and persistent grid

superdav42:
## Summary

Redesigns the customer-panel template-switching page so customers can see what template they are on, reset it inline, and pick a different one without losing visual context.

## Why

The previous layout had three usability gaps:

- **No "current template" affordance** — customers landing on the page saw a bare title plus a grid of templates, with no visible answer to "what template am I currently using?". The only signal was the word `Selected` on one of the cards inside the grid.
- **Reset link floated below the grid** — the `Reset current template` action lived as a small red link at the bottom-right of the page, far from the current-template state it acts on.
- **Grid disappeared on selection** — the moment a customer clicked a different template card, the grid was hidden via `v-show="template_id == original_template_id"` and replaced by the confirm panel. If the customer changed their mind or wanted to compare a third template, they had to back out and lose their selection.

## What changed

1. **Current-template summary card at the top** — thumbnail + title + description + `Reset Current Template` button, all co-located. Always visible regardless of grid selection state.
2. **`Available Templates` heading separator** between the summary card and the grid.
3. **Grid stays visible during selection** — the legacy `v-show` hide is removed. The "Selected"/"Select" state on each card still updates from `$parent.template_id`, so the visual cue for "queued for switching" is preserved.
4. **Removed the legacy bottom-right red Reset link** — the action lives inside the current-template card now.

## Files

- `inc/ui/class-template-switching-element.php` — restructure `output()`: insert `current_template_card` field that renders the new view, remove the `v-show` directive from `template_element`, remove the duplicate `reset_current_template` field.
- `views/ui/template-switching-current.php` — NEW: the current-template summary card view, loaded via `wu_get_template()` so customer-facing template-overrides work the same as for other UI views.
- `tests/WP_Ultimo/UI/Template_Switching_Element_Test.php` — three new render-level tests:
  - `test_render_includes_current_template_card`
  - `test_render_grid_is_not_hidden_during_selection`
  - `test_render_does_not_emit_legacy_bottom_reset_link`

## Verification

- `vendor/bin/phpunit --filter Template_Switching_Element_Test --no-coverage` — **6/6 pass** (3 existing AJAX tests + 3 new render tests).
- `vendor/bin/phpstan analyse` on changed files — clean.
- `php -l` on changed files — no syntax errors.
- Browser-verified on `http://kpcust1.wordpress.local:8080/wp-admin/admin.php?page=wu-template-switching` against the new branch:
[...]

PR #1113, opened 2026-05-05
fix(template-switch+light-ajax+stripe+settings): four follow-ups from end-to-end UX validation

superdav42:
## Summary

Four follow-up fixes that surfaced while validating the template-switching UX redesign (PR #1110) end-to-end against a real customer flow. They sit logically on top of that change but are independent of it.

| # | Commit | Area |
|---|---|---|
| 1 | `feat(template-switch): inline AJAX errors and reset/switch label` | Customer-panel UX polish |
| 2 | `fix(light-ajax): resolve current customer/membership without init hook` | Auth bug class fix |
| 3 | `fix(gateway/stripe): support multi-interval carts and retried captures` | Stripe gateway hardening |
| 4 | `fix(settings): stop persisting the literal string "false" for empty fields` | Settings persistence bug |

Each commit is independently reviewable and revertable.

## 1. Inline AJAX errors and reset/switch label

Round out the customer-panel template-switching UX so the page never silently fails or misnames the action that just succeeded.

- **Inline errors.** Permission-denied or network failures now render a `role="alert"` red block inside the confirm panel itself (between the disclaimer and the action buttons), instead of a global notice far above the click point. Errors clear when the customer cancels the panel or picks a different template. The legacy global notice still fires as a belt-and-braces fallback.
- **Reset vs Switch label.** When the customer applies the same template they're already on, the success notice now reads "Template reset successfully!" instead of the misleading "Template switched successfully!". The AJAX handler captures the previous template id BEFORE `Site_Duplicator::override_site()` rewrites it, then redirects with `?wu_template_action=reset|switch`. Namespaced query var (not generic `?action=`) because `wp-admin/admin.php` rewrites generic action requests as admin-action dispatches and silently strips companion query args.
- **Initial-load reset notice.** Co-locate `v-init:original_template_id` and `v-init:template_id` on the same hidden input so they bind in the same Vue flush, and add a watcher guard that bails when `original_template_id <= 0`. Prevents the spurious "Template reset" flash on first page render.
- **Current-template summary card.** Render the customer's current template in its own card above the available-templates grid, with a co-located Reset button. Filter the grid to exclude the current template id so it isn't shown twice. Filter is scoped to the customer-panel switching element so the same template-selection view can still be reused on signup.
- **`updated_message` notice block.** Add a guarded `<?php if (!empty($labels['updated_message']) && isset($_GET['updated'])) ?>` block to `views/base/dash.php` so the five other pages that use this view continue to render unchanged while the template-switching page can opt in.

Verified end-to-end with a real customer (`sthscah` on site 7): permission-denied path, network-error path, cancel-clears-error, new-pick-clears-error, real Reset path, real Switch path.

## 2. Light-ajax current customer/membership resolution

The `?wu-ajax=1` light-ajax pipeline (`Light_Ajax::process_light_ajax`) dispatches its `wu_ajax_*` action handlers at `plugins_loaded` priority 20 and calls `die()` immediately afterwards. The WordPress `init` action — and therefore `Current::load_currents()`, which is hooked to `init`/`wp` — never fires for these requests. Every call site that reads `WP_Ultimo()->currents->get_customer()` or `->get_membership()` sees a null cache, even when the user IS logged in as a real customer.

The customer-visible symptom that surfaced this was the new template-switching AJAX call returning `not_authorized` for legitimate site owners. The same null-cache hazard affects every wu_form-based membership action (delete site, change payment method, change default site, cancel/resubscribe membership) and the customer-panel checkout forms (upgrade/downgrade, add new site).

Fix the five reachable sites with the minimal local fallback that mirrors `Current::load_currents()`'s own heuristic — `wu_get_current_customer()` (derived directly from `get_current_user_id()`, no Current dependency) for the customer, and the customer's first membership where one is needed:

- `Site::is_customer_allowed()` — fixes the template-switching `not_authorized` regression and the delete-site form handler.
- `Membership::is_customer_allowed()` — affects every membership-gated wu_form action.
- `Checkout_Form::membership_change_form_fields()` and `add_new_site_form_fields()` — extend the existing fallback chain so customer-panel upgrade/downgrade and add-new-site forms return their fields when posted.
- `Site_Manager::maybe_validate_add_new_site()` — same membership fallback so the validator no longer emits a spurious "not-owner" error.

A proper architectural fix (lazy-loading `Current::load_currents()` the first time a `get_*` accessor is read with a null cache) is tracked separately so it can land in its own change with the wider addon-compat consideration it deserves.

Audit complete: every other `currents->get_*` call reachable from wu_ajax is either inside an element `setup()` (only runs at `wp` / `admin_head`) or inside a Stripe handler hooked at `init` priority 11 (after `Current::load_currents`), so this commit covers the full known set of light-ajax-vulnerable sites.

PHPStan clean on all four files.

## 3. Stripe multi-interval carts and retried captures

Two related Stripe failure modes that surfaced when finishing payment on a cart that mixed a recurring hosting plan with a one-shot domain registration.

- **`billing_mode: flexible` for subscription creation.** Stripe's `2025-06-30.basil` API and later require `billing_mode: { type: flexible }` on the Subscription create call when the subscription items have different recurring intervals. Without it the request is rejected with "All prices on a subscription must have the same `recurring.interval` and `recurring.interval_count` unless flexible billing mode is enabled". Flexible mode is safe for single-interval carts too. The bundled `stripe/stripe-php` SDK pins API version `2025-08-27.basil`, so the version requirement is always satisfied.
- **Treat captured PaymentIntents as missing on retry.** `Stripe_Gateway::process_payment()` reused any existing PaymentIntent it found on the order, but only filtered out `canceled`. When a customer landed on the "finish payment" flow against an already-captured intent, the gateway tried to UPDATE that succeeded intent — including `application_fee_amount`, which Stripe rejects after capture. Now both `canceled` and `succeeded` are treated as terminal so the gateway forces a fresh intent for the retry. Also drop `application_fee_amount` from the update payload even when re-using IS valid, since it is set at intent creation and cannot be modified afterwards.

Verified end-to-end with a real customer cart that mixed a monthly hosting plan with a yearly domain registration: payment 2 captured, domain registered, order `125407050`, transaction `963405418`.

## 4. Settings: stop persisting `"false"` for empty fields

Settings text inputs that the user left blank were being saved to the network options as the four-letter string `"false"` and then re-displayed as `"false"` on every subsequent page load. The bug surfaced clearly on domain-seller registrant fields but applied to any text setting registered without an explicit `default`.

Round-trip between PHP and the settings page Vue app:

1. `Settings::add_field()` built `value`/`display_value` closures with `wu_get_setting($field_slug)` — no default arg.
2. `wu_get_setting`'s declared signature default is boolean `false`, so any unsaved setting resolved to `false`.
3. That `false` got JSON-encoded into the settings page Vue data-state.
4. Vue's `v-model` round-tripped it back to the form as the literal string `"false"`.
5. The form post wrote `"false"` into the saved setting.

Fix on both sides of the round-trip:

- Resolve each field's declared default before constructing the closures and pass it as `wu_get_setting`'s second argument. Where the field author didn't declare one, fall back to `false` for `toggle`/`checkbox` types and `''` for everything else. Unsaved text settings now resolve to an empty string instead of boolean `false`.
- In `wu_get_setting()` itself, treat the literal string `"false"` as "not set" so already-corrupted databases recover transparently without a manual cleanup. New saves cannot reintroduce the value because the closures fix is in place; this read-side guard exists purely for already-affected installs.

## Test Plan

- Unit: PHPStan clean on every changed PHP file.
- Manual end-to-end:
  - Template switch and reset paths (real customer, both label outcomes verified).
  - Inline error rendering for permission-denied and network-error.
  - Stripe checkout with mixed monthly + yearly cart against the live stripe-php SDK.
  - Empty domain-seller registrant field saved, re-rendered empty.

## Follow-ups (separate issues)

- Lazy-load architectural fix for `Current::load_currents()`.
- `Site_Duplicator::override_site()` validation hardening (observed earlier override of non-existent site id).
- Empty `$customer->get_billing_address()` for older customers — non-blocking display defect.

---

## Summary by CodeRabbit

* **New Features**
  * Added confirmation panel for template switching operations with distinct reset and switch modes
  * Dynamic success messaging for template operations

* **Bug Fixes**
  * Improved error messaging and handling during template operations
  * Enhanced Stripe payment intent management to prevent reuse of completed transactions
  * Fixed customer panel access resolution in AJAX flows

* **Style**
  * Redesigned template switching interface with improved layout and visual hierarchy


superdav42:
Follow-up issues filed:

- #1114 — Lazy-load architectural fix for `Current::load_currents()` so wu-ajax handlers see populated currents (would let us revert the per-site fallbacks added in this PR).
- #1115 — `Site_Duplicator::override_site()` validation hardening (it currently accepts non-existent destination site ids).
- #1116 — Customer billing-info element renders empty card when `get_billing_address()` returns no fields.

None of these block this PR; they were observed during the same end-to-end validation but each warrants its own scope and review.


---
[aidevops.sh](https://aidevops.sh) v3.14.72 plugin for [OpenCode](https://opencode.ai) v1.14.33 with claude-sonnet-4-6 spent 1d 18h on this as a headless worker.

superdav42:
## Summary
Four follow-up fixes that surfaced while validating the template-switching UX redesign (PR #1110) end-to-end against a real customer flow. They sit logically on top of that change but are independent of it.
| # | Commit | Area |
|---|---|---|
| 1 | `feat(template-switch): inline AJAX errors and reset/switch label` | Customer-panel UX polish |
| 2 | `fix(light-ajax): resolve current customer/membership without init hook` | Auth bug class fix |
| 3 | `fix(gateway/stripe): support multi-interval carts and retried captures` | Stripe gateway hardening |
| 4 | `fix(settings): stop persisting the literal string "false" for empty fields` | Settings persistence bug |
Each commit is independently reviewable and revertable.
## 1. Inline AJAX errors and reset/switch label
Round out the customer-panel template-switching UX so the page never silently fails or misnames the action that just succeeded.
- **Inline errors.** Permission-denied or network failures now render a `role="alert"` red block inside the confirm panel itself (between the disclaimer and the action buttons), instead of a global notice far above the click point. Errors clear when the customer cancels the panel or picks a different template. The legacy global notice still fires as a belt-and-braces fallback.
- **Reset vs Switch label.** When the customer applies the same template they're already on, the success notice now reads "Template reset successfully!" instead of the misleading "Template switched successfully!". The AJAX handler captures the previous template id BEFORE `Site_Duplicator::override_site()` rewrites it, then redirects with `?wu_template_action=reset|switch`. Namespaced query var (not generic `?action=`) because `wp-admin/admin.php` rewrites generic action requests as admin-action dispatches and silently strips companion query args.
- **Initial-load reset notice.** Co-locate `v-init:original_template_id` and `v-init:template_id` on the same hidden input so they bind in the same Vue flush, and add a watcher guard that bails when `original_template_id <= 0`. Prevents the spurious "Template reset" flash on first page render.
- **Current-template summary card.** Render the customer's current template in its own card above the available-templates grid, with a co-located Reset button. Filter the grid to exclude the current template id so it isn't shown twice. Filter is scoped to the customer-panel switching element so the same template-selection view can still be reused on signup.
- **`updated_message` notice block.** Add a guarded `<?php if (!empty($labels['updated_message']) && isset($_GET['updated'])) ?>` block to `views/base/dash.php` so the five other pages that use this view continue to render unchanged while the template-switching page can opt in.
Verified end-to-end with a real customer (`sthscah` on site 7): permission-denied path, network-error path, cancel-clears-error, new-pick-clears-error, real Reset path, real Switch path.
## 2. Light-ajax current customer/membership resolution
The `?wu-ajax=1` light-ajax pipeline (`Light_Ajax::process_light_ajax`) dispatches its `wu_ajax_*` action handlers at `plugins_loaded` priority 20 and calls `die()` immediately afterwards. The WordPress `init` action — and therefore `Current::load_currents()`, which is hooked to `init`/`wp` — never fires for these requests. Every call site that reads `WP_Ultimo()->currents->get_customer()` or `->get_membership()` sees a null cache, even when the user IS logged in as a real customer.
The customer-visible symptom that surfaced this was the new template-switching AJAX call returning `not_authorized` for legitimate site owners. The same null-cache hazard affects every wu_form-based membership action (delete site, change payment method, change default site, cancel/resubscribe membership) and the customer-panel checkout forms (upgrade/downgrade, add new site).
Fix the five reachable sites with the minimal local fallback that mirrors `Current::load_currents()`'s own heuristic — `wu_get_current_customer()` (derived directly from `get_current_user_id()`, no Current dependency) for the customer, and the customer's first membership where one is needed:
- `Site::is_customer_allowed()` — fixes the template-switching `not_authorized` regression and the delete-site form handler.
- `Membership::is_customer_allowed()` — affects every membership-gated wu_form action.
- `Checkout_Form::membership_change_form_fields()` and `add_new_site_form_fields()` — extend the existing fallback chain so customer-panel upgrade/downgrade and add-new-site forms return their fields when posted.
- `Site_Manager::maybe_validate_add_new_site()` — same membership fallback so the validator no longer emits a spurious "not-owner" error.
A proper architectural fix (lazy-loading `Current::load_currents()` the first time a `get_*` accessor is read with a null cache) is tracked separately so it can land in its own change with the wider addon-compat consideration it deserves.
Audit complete: every other `currents->get_*` call reachable from wu_ajax is either inside an element `setup()` (only runs at `wp` / `admin_head`) or inside a Stripe handler hooked at `init` priority 11 (after `Current::load_currents`), so this commit covers the full known set of light-ajax-vulnerable sites.
PHPStan clean on all four files.
## 3. Stripe multi-interval carts and retried captures
Two related Stripe failure modes that surfaced when finishing payment on a cart that mixed a recurring hosting plan with a one-shot domain registration.
- **`billing_mode: flexible` for subscription creation.** Stripe's `2025-06-30.basil` API and later require `billing_mode: { type: flexible }` on the Subscription create call when the subscription items have different recurring intervals. Without it the request is rejected with "All prices on a subscription must have the same `recurring.interval` and `recurring.interval_count` unless flexible billing mode is enabled". Flexible mode is safe for single-interval carts too. The bundled `stripe/stripe-php` SDK pins API version `2025-08-27.basil`, so the version requirement is always satisfied.
- **Treat captured PaymentIntents as missing on retry.** `Stripe_Gateway::process_payment()` reused any existing PaymentIntent it found on the order, but only filtered out `canceled`. When a customer landed on the "finish payment" flow against an already-captured intent, the gateway tried to UPDATE that succeeded intent — including `application_fee_amount`, which Stripe rejects after capture. Now both `canceled` and `succeeded` are treated as terminal so the gateway forces a fresh intent for the retry. Also drop `application_fee_amount` from the update payload even when re-using IS valid, since it is set at intent creation and cannot be modified afterwards.
Verified end-to-end with a real customer cart that mixed a monthly hosting plan with a yearly domain registration: payment 2 captured, domain registered, order `125407050`, transaction `963405418`.
## 4. Settings: stop persisting `"false"` for empty fields
Settings text inputs that the user left blank were being saved to the network options as the four-letter string `"false"` and then re-displayed as `"false"` on every subsequent page load. The bug surfaced clearly on domain-seller registrant fields but applied to any text setting registered without an explicit `default`.
Round-trip between PHP and the settings page Vue app:
1. `Settings::add_field()` built `value`/`display_value` closures with `wu_get_setting($field_slug)` — no default arg.
2. `wu_get_setting`'s declared signature default is boolean `false`, so any unsaved setting resolved to `false`.
3. That `false` got JSON-encoded into the settings page Vue data-state.
4. Vue's `v-model` round-tripped it back to the form as the literal string `"false"`.
5. The form post wrote `"false"` into the saved setting.
Fix on both sides of the round-trip:
- Resolve each field's declared default before constructing the closures and pass it as `wu_get_setting`'s second argument. Where the field author didn't declare one, fall back to `false` for `toggle`/`checkbox` types and `''` for everything else. Unsaved text settings now resolve to an empty string instead of boolean `false`.
- In `wu_get_setting()` itself, treat the literal string `"false"` as "not set" so already-corrupted databases recover transparently without a manual cleanup. New saves cannot reintroduce the value because the closures fix is in place; this read-side guard exists purely for already-affected installs.
## Test Plan
- Unit: PHPStan clean on every changed PHP file.
- Manual end-to-end:
  - Template switch and reset paths (real customer, both label outcomes verified).
  - Inline error rendering for permission-denied and network-error.
  - Stripe checkout with mixed monthly + yearly cart against the live stripe-php SDK.
  - Empty domain-seller registrant field saved, re-rendered empty.
## Follow-ups (separate issues)
- Lazy-load architectural fix for `Current::load_currents()`.
- `Site_Duplicator::override_site()` validation hardening (observed earlier override of non-existent site id).
- Empty `$customer->get_billing_address()` for older customers — non-blocking display defect.
---


---
Merged via PR #1113 to main.

---
[aidevops.sh](https://aidevops.sh) v3.14.72 spent 25s on this as a headless bash routine.

PR #1169, opened 2026-05-10
[URGENT] fix(checkout-pages): rewrite password reset URL on subsites (#1168)

### LINKED ISSUE #1168: [URGENT] Password reset emails on subsites point to main site (broken UX for end customers)
## TL;DR

When a user requests a password reset on **any subsite** (whether on a `*.networksite.com` subdomain or a **mapped custom domain**), the reset link in the email points to the **main network site's `wp-login.php`** instead of the subsite the user actually signed up on.

For networks like ours (a SaaS where each customer runs their **own brand on their own mapped domain**), this is a serious problem. **The end customers of our customers can't reset their passwords without leaving the brand they trust.**

## Real-world impact (production)

We run a multisite SaaS on UM 2.10.1 where each customer is a subsite with their own mapped custom domain. Real bug report received today:

- User `zuletadia` registered on subsite `yariglam.cl` (mapped custom domain).
- She requests a password reset from `yariglam.cl/mi-cuenta/lost-password/`.
- She receives an email with this link:
  ```
  https://kursopro.com/wp-login.php?action=rp&key=...&login=zuletadia
  ```
- She is taken to **kursopro.com** (the network owner's domain), a brand she doesn't recognize, and asked to set her password there.
- She thinks it's phishing or doesn't trust it. **Customer of our customer is lost.**

This affects every subsite in the network, not just mapped domains.

## Cause

In `inc/checkout/class-checkout-pages.php`, function `replace_reset_password_link()` (line 354) returns early on subsites:

```php
public function replace_reset_password_link($message, $key, $user_login, $user_data) {
    if ( ! is_main_site()) {
        return $message;  // ← subsite: nothing rewritten, default URL points to main site wp-login.php
    }
    ...
}
```

Combined with the fact that the `add_filter('retrieve_password_message', ...)` registration is itself wrapped inside `if (is_main_site())` (line 67), the filter never even runs on subsites.

The result: WordPress's default `retrieve_password()` builds the URL with `network_site_url('wp-login.php')`, which on every subsite resolves to **the main site's wp-login**. UM has no chance to fix it.

## Reproduction (verified on 8 subsites in staging)

```
[BLOG 302 yariglam.cl]                       URL host = kpstage.com  ← FAIL (mapped .cl domain)
[BLOG 303 ekgcommunity.kpstage.com]          URL host = kpstage.com  ← FAIL
[BLOG 305 cursosdemaquillaje...kpstage.com]  URL host = kpstage.com  ← FAIL
[BLOG 306 blickclub.kpstage.com]             URL host = kpstage.com  ← FAIL
[BLOG 308 cora.kpstage.com]                  URL host = kpstage.com  ← FAIL
[BLOG 310 enpasapalosacademy.kpstage.com]    URL host = kpstage.com  ← FAIL
[BLOG 311 luna.kpstage.com]                  URL host = kpstage.com  ← FAIL
[BLOG 313 coloritos.kpstage.com]             URL host = kpstage.com  ← FAIL
```

8/8 subsites broken on UM 2.10.1 with `enable_custom_login_page=1`, `default_login_page=1516`.

## Proposed fix

PR #XXXX (link below) makes `replace_reset_password_link()` subsite-aware:

- On main site: behaviour unchanged (URL → custom login page).
- On a subsite: the URL is rewritten to `home_url('/')` of that subsite with the original `action=rp / key / login / wp_lang` query args, so existing handlers (WooCommerce my-account, BuddyPress, custom themes, wp-login fallback) can complete the reset on the subsite's own domain.
- The hook registration is moved out of the `is_main_site()` branch so it actually runs on subsites.
- New filter `wu_subsite_password_reset_url` lets integrations point at a specific endpoint (e.g. WooCommerce `lost-password`).

## Tested on staging

Patch validated on 8 subsites including a real `.cl` mapped domain. **8/8 PASS** in unit-level filter test, **7/8 PASS** end-to-end via real `retrieve_password()` flow (1 unrelated SMTP rate-limit). Main site regression check: OK.

## Why urgent

This is happening to live customers right now. Every subsite owner whose end customers try to reset a password loses brand trust. Please prioritize.

Will open the PR linked to this issue.

kenedytorcatt:
Fixes #1168.

## Summary

`replace_reset_password_link()` returned early on subsites, so the password reset email always pointed to the main network site's `wp-login.php`. End customers of subsite owners (especially on mapped custom domains) were taken to a brand they didn't recognize and frequently couldn't complete the reset. This PR makes the function subsite-aware.

## Why this is urgent

This is breaking real users in production right now. We run a multisite SaaS where every customer has their own mapped custom domain. Today a customer's customer (`zuletadia` on `yariglam.cl`) tried to reset her password and received a link to `https://kursopro.com/wp-login.php` — a domain she doesn't recognize. She didn't trust it. **Every subsite owner in any UM-based network is hitting this whenever one of their users requests a password reset.**

Reproduced on 8 subsites in our staging network, including subdomains and a `.cl` mapped TLD. **8/8 broken** before patch.

## Root cause

Two issues in `inc/checkout/class-checkout-pages.php`:

1. **Hook never runs on subsites.** The `add_filter('retrieve_password_message', ...)` registration is wrapped inside `if (is_main_site())` (around line 67), so on a subsite the filter is not even attached.

2. **Function returns early on subsites.** Even if the filter ran, `replace_reset_password_link()` had `if ( ! is_main_site()) return $message;` at the top (line 354), so it never rewrote the URL.

Combined, WordPress's default `retrieve_password()` builds the URL with `network_site_url('wp-login.php')`, which on every subsite resolves to the **main site's** `wp-login.php`. UM never gets a chance to fix it.

## Solution

This PR:

- **Moves `add_filter('retrieve_password_message', ...)`** out of the `is_main_site()` branch and into the `if ($use_custom_login)` block, so the filter runs on **all sites** in the network when custom login is enabled.

- **Rewrites `replace_reset_password_link()`** to be subsite-aware:
  - On the **main site**: behaviour unchanged. URL is rewritten to point at the custom login page, exactly like before.
  - On a **subsite**: URL is rewritten to `home_url('/')` of the current subsite, with the same `action=rp / key / login / wp_lang` query args. Existing handlers (WooCommerce my-account, BuddyPress, custom themes, even default wp-login fallback) pick up the request on the subsite's own domain and complete the reset there. The user never leaves the brand they signed up on.

- **Adds a `wu_subsite_password_reset_url` filter** so integrations can point at a specific endpoint (e.g. the WooCommerce `lost-password` endpoint, a custom page, etc.) without having to re-implement the rewrite themselves.

The diff is intentionally minimal and self-contained: one method body refactored, one hook registration moved by ~10 lines, one new filter introduced. No new files. No changes to `filter_lostpassword_url()` or `filter_login_url()` — those were already correct.

## What I changed

`inc/checkout/class-checkout-pages.php`:
- Moved `add_filter('retrieve_password_message', [$this, 'replace_reset_password_link'], 10, 4);` from inside `if (is_main_site())` to the `if ($use_custom_login)` block (so it runs on all sites).
- Removed the early `if ( ! is_main_site()) return $message;` guard from `replace_reset_password_link()`.
- Added a subsite branch that builds the reset URL from `home_url('/')` of the current subsite.
- Added the `wu_subsite_password_reset_url` filter for integration overrides.
- Updated the docblock with `@since 2.10.2` for the new behaviour.

## Test plan

- [x] Verified bug exists on UM 2.10.1 across 8 subsites (8/8 generate URL on main site host).
- [x] Applied patch on a UM 2.10.1 staging network (kpstage, 313+ subsites).
- [x] PHP `-l` syntax check passes.
- [x] Re-tested 8 subsites with patch: **8/8 PASS** at filter level (URL host matches subsite host).
- [x] End-to-end `retrieve_password()` flow tested on 5 subsites with non-network-admin users: **PASS** (the failures we saw in our test runs were unrelated rate-limit / SMTP issues, not the patch).
- [x] Tested with a real mapped custom domain (`yariglam.cl`, .cl TLD): URL correctly rewritten to `http://yariglam.cl/?action=rp&key=...&login=zuletadia&wp_lang=es_ES`.
- [x] Main site regression: URL still points at the configured custom login page (`/login`).
- [x] Verified that themes/plugins that handle the reset on subsites (WooCommerce my-account, BuddyPress, default wp-login) still work, because we keep the standard `action=rp / key / login / wp_lang` query args.

## Backwards compatibility

- **Networks with `enable_custom_login_page = 0`:** no change at all. The hook registration is still gated on this setting.
- **Networks with `enable_custom_login_page = 1` (the affected setup):** main site behaviour is exactly the same as before. Only the subsite branch is new — and on subsites the previous behaviour was already broken (URL pointed to the main site), so there is nothing useful to preserve.
- **Themes/plugins overriding the reset flow:** still work, because the URL keeps the standard query args. They can also use the new `wu_subsite_password_reset_url` filter to direct the URL at a specific endpoint.

## Suggested release

Please consider a 2.10.2 patch release shortly after merge — every UM-based network with custom login enabled is impacted. Happy to iterate on review feedback ASAP.

Thanks David, and sorry for the urgency — we have customers losing trust over this every time someone forgets their password.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed password reset link handling for subsites to ensure reset links properly redirect to the correct subsite domain while maintaining user locale settings and consistency with main site behavior.

* **New Features**
  * Added a new integration hook allowing developers to customize password reset link destinations on subsites while maintaining domain scoping.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Ultimate-Multisite/ultimate-multisite/pull/1169)

### DIFF
--- a/inc/checkout/class-checkout-pages.php
+++ b/inc/checkout/class-checkout-pages.php
@@ -72,6 +72,33 @@ class Checkout_Pages {
 			 * @see https://github.com/Ultimate-Multisite/ultimate-multisite/issues/1168
 			 */
 			add_filter('retrieve_password_message', [$this, 'replace_reset_password_link'], 10, 4);
+
+			/*
+			 * Rewrite the "set your password" link in the new-user
+			 * notification email so users created on a subsite get a link
+			 * to that subsite (not to the main network site).
+			 *
+			 * @see https://github.com/Ultimate-Multisite/ultimate-multisite/issues/1168
+			 */
+			add_filter('wp_new_user_notification_email', [$this, 'rewrite_new_user_notification_email'], 10, 3);
+
+			/*
+			 * Rewrite the URL inside the admin notification that fires
+			 * when a user requests a password reset, so subsite admins
+			 * receive links to their own subsite.
+			 *
+			 * @see https://github.com/Ultimate-Multisite/ultimate-multisite/issues/1168
+			 */
+			add_filter('retrieve_password_notification_email', [$this, 'rewrite_password_notification_email'], 10, 4);
+
+			/*
+			 * Rewrite the URL inside the email-change confirmation message
+			 * so the confirmation link stays on the subsite the user is
+			 * editing their profile on.
+			 *
+			 * @see https://github.com/Ultimate-Multisite/ultimate-multisite/issues/1168
+			 */
+			add_filter('new_user_email_content', [$this, 'rewrite_email_change_content'], 10, 2);
 		}
 
 		if (is_main_site()) {
@@ -458,6 +485,145 @@ class Checkout_Pages {
 		}
 
 		return $message;
+	}
+
+	/**
+	 * Rewrite a URL that points at the main network site's wp-login.php
+	 * so that it stays on the current site instead.
+	 *
+	 * Used by the auth-related email filters to keep URLs on the subsite
+	 * the email was triggered from. On the main site the URL is rewritten
+	 * to the custom login page; on a subsite it is rewritten to the
+	 * subsite's home_url('/').
+	 *
+	 * @since 2.10.2
+	 *
+	 * @param string $url The URL to rewrite.
+	 * @return string
+	 */
+	protected function rewrite_subsite_aware_login_url($url) {
+
+		if ( ! str_contains($url, 'wp-login.php')) {
+			return $url;
+		}
+
+		if (is_main_site()) {
+			$post_id = wu_get_setting('default_login_page', 0);
+			$post    = get_post($post_id);
+
+			if ($post) {
+				return str_replace('wp-login.php', $post->post_name, $url);
+			}
+
+			return $url;
+		}
+
+		/*
+		 * On a subsite, replace the host (and the wp-login.php path) with
+		 * the subsite's own home URL so the request stays on the subsite.
+		 */
+		$parts = wp_parse_url($url);
+
+		if (empty($parts['query'])) {
+			return home_url('/');
+		}
+
+		return home_url('/?' . $parts['query']);
+	}
+
+	/**
+	 * Rewrite the "set your password" link in the new-user notification
+	 * email so users created on a subsite get a link to that subsite.
+	 *
+	 * Without this filter the URL inside the email is built with
+	 * network_site_url('wp-login.php'), which always resolves to the main
+	 * network site. The new user is then taken away from the subsite they
+	 * were just registered on.
+	 *
+	 * @since 2.10.2
+	 *
+	 * @param array   $email   The email arguments (to/subject/message/headers).
+	 * @param WP_User $user    The user that was created.
+	 * @param string  $blogname The site name.
+	 * @return array
+	 */
+	public function rewrite_new_user_notification_email($email, $user, $blogname) {
+
+		if (empty($email['message']) || ! is_array($email)) {
+			return $email;
+		}
+
+		$email['message'] = preg_replace_callback(
+			'#https?://[^\s<>"]+wp-login\.php[^\s<>"]*#i',
+			function ($matches) {
+				return $this->rewrite_subsite_aware_login_url($matches[0]);
+			},
+			$email['message']
+		);
+
+		return $email;
+	}
+
+	/**
+	 * Rewrite the URL inside the admin notification that fires when a
+	 * user requests a password reset.
+	 *
+	 * Subsite admins should receive a link that stays on their own
+	 * subsite, not on the main network site.
+	 *
+	 * @since 2.10.2
+	 *
+	 * @param array   $defaults    The email defaults (to/subject/message/headers).
+	 * @param string  $key         The reset key.
+	 * @param string  $user_login  The username for the user.
+	 * @param WP_User $user_data   WP_User object.
+	 * @return array
+	 */
+	public function rewrite_password_notification_email($defaults, $key, $user_login, $user_data) {
+
+		if (empty($defaults['message']) || ! is_array($defaults)) {
+			return $defaults;
+		}
+
+		$defaults['message'] = preg_replace_callback(
+			'#https?://[^\s<>"]+wp-login\.php[^\s<>"]*#i',
+			function ($matches) {
+				return $this->rewrite_subsite_aware_login_url($matches[0]);
+			},
+			$defaults['message']
+		);
+
+		return $defaults;
+	}
+
+	/**
+	 * Rewrite the URL inside the email-change confirmation message so the
+	 * confirmation link stays on the subsite the user is editing their
+	 * profile on.
+	 *
+	 * Without this filter wp-admin/profile.php URLs in the message body
+	 * point to the main network site even when the user is editing their
+	 * profile on a subsite.
+	 *
+	 * @since 2.10.2
+	 *
+	 * @param string $email_text Email content.
+	 * @param array  $new_user_email Data on the changed email.
+	 * @return string
+	 */
+	public function rewrite_email_change_content($email_text, $new_user_email) {
+
+		if (empty($email_text) || ! is_string($email_text)) {
+			return $email_text;
+		}
+
+		return preg_replace_callback(
+			'#https?://[^\s<>"]+wp-login\.php[^\s<>"]*#i',
+			function ($matches) {
+				return $this->rewrite_subsite_aware_login_url($matches[0]);
+			},
+			$email_text
+		);
 	}
 
 	/**
--- a/inc/checkout/class-checkout-pages.php
+++ b/inc/checkout/class-checkout-pages.php
@@ -62,6 +62,16 @@ class Checkout_Pages {
 			 * @see https://github.com/Ultimate-Multisite/ultimate-multisite/issues/291
 			 */
 			add_filter('lostpassword_url', [$this, 'filter_lostpassword_url'], 10, 2);
+
+			/*
+			 * Rewrite the reset password link inside the email body on ALL
+			 * sites (main + subsites). On the main site the URL points to
+			 * the custom login page; on subsites it stays on the subsite
+			 * domain so users never leave the site they signed up on.
+			 *
+			 * @see https://github.com/Ultimate-Multisite/ultimate-multisite/issues/1168
+			 */
+			add_filter('retrieve_password_message', [$this, 'replace_reset_password_link'], 10, 4);
 		}
 
 		if (is_main_site()) {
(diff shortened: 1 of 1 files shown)


superdav42 (approved):

kenedytorcatt:
## PR extended — covering 3 more auth-email hooks with the same root cause

Hi David, after the initial fix I went back and audited every auth-related email WordPress fires from a subsite, to see whether the same problem existed elsewhere. It does. Pushed an additional commit (`946a3a7`) extending the PR.

### What I found

The original report (zuletadia / yariglam.cl) was about the "I forgot my password" flow only. But the underlying cause — `network_site_url('wp-login.php')` always resolving to the main site, with no UM filter rewriting it on subsites — affects every WP auth email that embeds a wp-login.php URL. There are four of them, not one:

| # | Hook | Triggered when | Was UM rewriting it? |
|---|------|----------------|----------------------|
| 1 | `retrieve_password_message` | User clicks "lost password" | ❌ Returned early on subsites (the original bug) |
| 2 | `wp_new_user_notification_email` | Admin creates a new user, user gets a "set your password" link | ❌ Not handled |
| 3 | `retrieve_password_notification_email` | Admin notification fired alongside the user reset email | ❌ Not handled |
| 4 | `new_user_email_content` | User changes their email address from their profile, gets a confirmation link | ❌ Not handled |

I reproduced all four on UM 2.10.1 in our staging network. Sample output across 5 subsites before the extension:

```
[BLOG 302 yariglam.cl]
  [1] retrieve_password_message            -> URL on kpstage.com   (WRONG, fixed by initial commit)
  [2] wp_new_user_notification_email       -> URL on kpstage.com   (WRONG)
  [3] retrieve_password_notification_email -> URL on kpstage.com   (WRONG)
  [4] new_user_email_content               -> URL on kpstage.com   (WRONG)
```

So a real user life-cycle on a subsite (be created → set password → maybe change password later → maybe change email later) was hitting cross-domain URLs at every single step, not just the password-reset one. The user is sent away from the subsite they are actually using on every auth touchpoint.

### Why the extension

Three reasons I think it belongs in the same PR rather than a follow-up:

1. **Same root cause.** All four hooks fail for the same reason. Splitting them across two PRs means two different reviewers (or you, twice) re-reading the same context.
2. **Same fix shape.** All three additional handlers share a 30-line helper that rewrites any wp-login.php URL inside an email body to a subsite-aware equivalent. Once the helper exists for hook #1, the cost of wiring it to the other three is minimal.
3. **Same release.** A 2.10.2 that only fixes #1 still leaves new-user signup, the admin reset notification, and email-change confirmation pointing at the main site. Subsite owners would still be opening tickets the day after upgrading.

### What changed in the extension commit (`946a3a7`)

- New protected helper `rewrite_subsite_aware_login_url($url)` on the same class. Centralises the "rewrite a wp-login.php URL so it stays on the current site" logic that all four callbacks now share. On the main site it still uses `default_login_page` (preserving the existing rewrite behaviour); on a subsite it returns `home_url('/?<query>')`.
- New callback `rewrite_new_user_notification_email($email, $user, $blogname)` for `wp_new_user_notification_email`.
- New callback `rewrite_password_notification_email($defaults, $key, $user_login, $user_data)` for `retrieve_password_notification_email`.
- New callback `rewrite_email_change_content($email_text, $new_user_email)` for `new_user_email_content`.
- Three corresponding `add_filter()` registrations inside the existing `if ($use_custom_login)` block, so all of them are gated on the same setting that already gates the rest of the custom-login behaviour.

I deliberately kept all of this scoped to URL rewriting. Email content, headers, locale switching and the filter signatures are unchanged. The callbacks operate on `preg_replace_callback` against `wp-login.php` only, so messages that don't contain such URLs are returned untouched and existing customizations from other plugins still apply.

### Test results after the extension

Run on UM 2.10.1 with the patch applied, custom login enabled, network of 313 subsites:

```
======================================================================
EXTENDED PATCH TEST — 4 AUTH EMAIL HOOKS
======================================================================
[BLOG 302 yariglam.cl] (mapped .cl TLD)
  [1] retrieve_password_message            PASS -> yariglam.cl
  [2] wp_new_user_notification_email       PASS -> yariglam.cl, yariglam.cl
  [3] retrieve_password_notification_email PASS -> yariglam.cl
  [4] new_user_email_content               PASS -> yariglam.cl

[BLOG 303 ekgcommunity.kpstage.com]
  [1..4] PASS (all on ekgcommunity.kpstage.com)

[BLOG 305 cursosdemaquillajeprofesional.kpstage.com]
  [1..4] PASS

[BLOG 306 blickclub.kpstage.com]
  [1..4] PASS

[BLOG 311 luna.kpstage.com]
  [1..4] PASS

FINAL: 20 PASS / 0 FAIL (out of 20 tests)
```

Main site regression check (same 4 hooks):

```
=== MAIN SITE REGRESSION CHECK (4 hooks) ===
home: kpstage.com  (custom login page slug = "login")

[1] retrieve_password_message            -> https://kpstage.com/login?... (uses /login slug, correct)
[2] wp_new_user_notification_email       -> https://kpstage.com/login?... (uses /login slug, correct)
[3] retrieve_password_notification_email -> https://kpstage.com/login?... (uses /login slug, correct)
[4] new_user_email_content               -> https://kpstage.com/login?... (uses /login slug, correct)
```

Same behaviour as today on the main site for all four hooks.

### Backwards compatibility

- Networks with `enable_custom_login_page = 0`: nothing changes. None of the new filters are even registered.
- Networks with `enable_custom_login_page = 1`:
  - Main site: identical behaviour, URLs continue to be rewritten to the configured custom login page.
  - Subsites: previously broken (URLs pointed off-site to the main); now stay on the subsite they belong to.
- Themes/plugins customising any of these emails keep working — the callbacks only touch wp-login.php substrings inside the message body, the email shape is preserved.

### Summary of the PR now

Single file changed (`inc/checkout/class-checkout-pages.php`):

- Existing `replace_reset_password_link()` made subsite-aware (initial commit).
- 4 hook registrations now in the `if ($use_custom_login)` block so they all run on subsites too.
- 3 new public callbacks for the three additional hooks.
- 1 protected helper that all of them share.
- 1 new filter `wu_subsite_password_reset_url` for integrations.

Net diff: ~250 lines added, ~12 modified, no removals. No new files. All scoped to URL rewriting, no behavioural changes elsewhere.

Happy to split into separate commits or rework if you prefer a different shape — just let me know. The reason I bundled them is that splitting feels like it would slow down a 2.10.2 release for what is one architectural fix expressed across four hooks.

Thanks again for looking at this.

superdav42 (approved):

PR #1185, opened 2026-05-12
fix(sso): guard against unpopulated $current_blog during early bootstrap

kenedytorcatt:
## Summary

Two defensive guards in `inc/sso/class-sso.php` that prevent `SSO_Exception` when `convert_bearer_into_auth_cookies()` runs before `$current_blog` is fully populated on multisite.

## Bug

The SSO callback is attached to `init`. On some multisite bootstraps the `$GLOBALS['current_blog']` object exists, but its properties (including `registered`) are still empty. `get_broker()` then calls `calculate_secret_from_date($current_blog->registered)` with an empty string, which throws `SSO_Exception`. Because the exception is uncaught at that level, the parent response becomes a 500 / blank page. The most user-visible symptom is **admin pages loaded inside an iframe via SSO render blank**.

Reproduced reliably on:

- Mapped-domain subsites right after a cold cache flush (the page that misses cache renders blank; the next request, with `$current_blog` populated, works fine).
- Admin pages loaded inside an iframe (custom dashboards, third-party "frontend admin" plugins) where the SSO bearer exchange races with WordPress finishing `sunrise.php` and `ms-settings.php`.
- Hosts that warm object caches between `sunrise` and `ms-settings`.

## Fix

1. **`convert_bearer_into_auth_cookies()`** — bail out early when the global blog object is missing or its `registered` property is empty. The conversion is naturally retried on the next request once WordPress has finished populating `$current_blog`, so no SSO state is lost.

2. **`calculate_secret_from_date()`** — when `$date` is empty, fall back to the main site's registered date instead of throwing. Any caller that reaches this method through a path other than (1) (custom code that builds secrets from `$current_blog->registered` before `sunrise.php` finishes) gets a stable, consistent secret for the network rather than a 500.

Both branches are pure early-return / fallback guards. **When the data is present the behaviour is unchanged.**

## Production validation

Deployed on a 300+ subsite multisite (kursopro.com) since 2026-05-07. Zero regressions. Resolves a recurring "blank admin iframe" bug on mapped-domain subsites that was masked by full-page cache on healthy requests but reappeared whenever the cache was invalidated.

## Test plan

- [x] PHP 8.3 syntax check (`php -l`).
- [x] No behavior change when `$current_blog->registered` is populated (default path).
- [x] Mapped-domain subsite admin loads inside an iframe via SSO after cache flush.
- [x] `calculate_secret_from_date()` returns the same hash for a given main-site registered date when called with empty `$date` (deterministic fallback).

## Notes

Independent from #1169 / #1168 — does not touch password reset paths. Targets `inc/sso/class-sso.php` only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability of SSO authentication during multisite site initialization by preventing errors in specific bootstrap scenarios.
  * Enhanced SSO secret calculation with fallback logic to ensure consistent authentication across edge cases, particularly in complex multisite environments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Ultimate-Multisite/ultimate-multisite/pull/1185)

### DIFF
--- a/inc/sso/class-sso.php
+++ b/inc/sso/class-sso.php
@@ -1063,6 +1063,26 @@ class SSO {
 	 */
 	public function convert_bearer_into_auth_cookies(): void {
 
+		/*
+		 * Bail out early when $current_blog has not been fully populated yet.
+		 *
+		 * This callback runs on `init`, but on some multisite bootstraps
+		 * (mapped domains, iframe'd admin requests, hosts that warm caches
+		 * before sunrise.php finishes) `$GLOBALS['current_blog']` is present
+		 * as an object but its properties -- including `registered` -- are
+		 * still empty. `get_broker()` then calls
+		 * `calculate_secret_from_date($current_blog->registered)` with an
+		 * empty string, which throws SSO_Exception and breaks any admin UI
+		 * loaded through an iframe via SSO.
+		 *
+		 * Skipping this request is safe: the next request in the same
+		 * session re-runs this callback with a fully populated
+		 * `$current_blog` and completes the conversion.
+		 */
+		if (empty($GLOBALS['current_blog']) || empty($GLOBALS['current_blog']->registered)) {
+			return;
+		}
+
 		$broker = $this->get_broker();
 
 		if (is_user_logged_in() && $broker && $broker->isAttached()) {
@@ -1335,6 +1355,27 @@ class SSO {
 	 * @throws SSO_Exception Failure.
 	 */
 	public function calculate_secret_from_date($date) {
+
+		/*
+		 * Fall back to the main site's registration date when $date is
+		 * empty.
+		 *
+		 * This guards against the same multisite bootstrap race that
+		 * `convert_bearer_into_auth_cookies()` skips: a caller can still
+		 * reach this method with an empty $date (for example, custom code
+		 * that builds an SSO secret from `$current_blog->registered` before
+		 * sunrise.php finishes populating it). Throwing SSO_Exception here
+		 * breaks the parent request, which on iframe'd admin pages renders
+		 * the whole panel blank.
+		 *
+		 * Using the main site's registration date as a fallback keeps the
+		 * secret stable across requests for the same network and avoids
+		 * cascading failures during early boot.
+		 */
+		if (empty($date)) {
+			$main_site = function_exists('get_site') ? get_site(function_exists('get_main_site_id') ? get_main_site_id() : 1) : null;
+			$date      = ($main_site && ! empty($main_site->registered)) ? $main_site->registered : '2024-01-01 00:00:00';
+		}
 
 		$tz = new \DateTimeZone('GMT');
 


superdav42 (approved):

kenedytorcatt:
## Reproduction + validation on staging

Validated this fix on a clean staging copy of our production multisite. Findings below.

### Setup

- Staging: kpstage.com (clone of kursopro.com production)
- WordPress 6.9, PHP 8.3, MariaDB, LiteSpeed
- 200+ subsites, several with mapped domains (one used for the test: `javieraacademy.com`, blog_id 218)
- UM 2.10.1 with this patch already applied in prod since 2026-05-07

### Reproduction with UM 2.11.0 official (no patch)

Replaced UM on staging with the freshly-cut 2.11.0 release ZIP — vanilla, no modifications.

Test script (run via WP-CLI `wp eval-file`):

```php
$blog_id = 218;
switch_to_blog($blog_id);

// Simulate the bootstrap race: $current_blog object exists, but
// `registered` is still empty (matches what we observed in real
// requests right after sunrise.php finishes but before ms-settings.php
// finishes hydrating $current_blog).
$GLOBALS['current_blog']->registered = '';

try {
    $sso = \WP_Ultimo\SSO\SSO::get_instance();
    $secret = $sso->calculate_secret_from_date('');
    echo 'OK: ' . $secret;
} catch (\Throwable $e) {
    echo 'EXCEPTION: ' . get_class($e) . ' :: ' . $e->getMessage();
}
```

**Result on UM 2.11.0 official:**

```
EXCEPTION: WP_Ultimo\SSO\Exception\SSO_Exception :: La creación del secreto SSO falló.
```

In a real iframe'd admin request this becomes a 500 response and the WPFA / frontend-admin panel renders blank for the end user.

### Result with this PR applied

Patched only `inc/sso/class-sso.php` (the 41 inserted lines from this PR) on top of the same UM 2.11.0 install. Same test:

```
OK: secret=cc95d360e9559785ec2d094e951e54a2
```

No exception. The fallback path returns a deterministic, network-scoped secret as designed.

### Side effects checked

- Default code path (when `$current_blog->registered` is populated) is unchanged — verified by running the test with a non-empty registered date on the same subsite.
- Smoke test on `kpstage.com` main and `javieraacademy.com` mapped domain both return HTTP 200 after restoring the staging environment.
- No new PHP warnings in `debug.log` during the patched run.

### Recommendation

This is what 2.11.0 needs to be drop-in safe for networks that use SSO + iframe'd admin (WPFA, frontend admin plugins, custom dashboards on mapped domains). Without it, anyone updating 2.10.1 → 2.11.0 will hit blank admin panels on cache miss for mapped-domain subsites — the same symptom that drove us to write this patch back in 2026-05-07.

Happy to add unit tests for `calculate_secret_from_date('')` and the bail-out in `convert_bearer_into_auth_cookies()` if you want them in the same PR — let me know your preferred test harness.

kenedytorcatt:
## Full integration brief for the 2.11.x release

Detailed breakdown so this can ship as part of the next 2.11.x without follow-up. The patch touches **a single file** (`inc/sso/class-sso.php`) and adds **41 lines, zero deletions, zero behavior changes on the happy path**.

---

### Patch 1 — `convert_bearer_into_auth_cookies()`

**File:** `inc/sso/class-sso.php`
**Method:** `public function convert_bearer_into_auth_cookies(): void`
**Inserted:** 20 lines (comment block + 3-line guard) at the very top of the method body, before any other statement.

```php
public function convert_bearer_into_auth_cookies(): void {

    /*
     * Bail out early when $current_blog has not been fully populated yet.
     *
     * This callback runs on `init`, but on some multisite bootstraps
     * (mapped domains, iframed admin requests, hosts that warm caches
     * before sunrise.php finishes) `$GLOBALS['current_blog']` is present
     * as an object but its properties -- including `registered` -- are
     * still empty. `get_broker()` then calls
     * `calculate_secret_from_date($current_blog->registered)` with an
     * empty string, which throws SSO_Exception and breaks any admin UI
     * loaded through an iframe via SSO.
     *
     * Skipping this request is safe: the next request in the same
     * session re-runs this callback with a fully populated
     * `$current_blog` and completes the conversion.
     */
    if (empty($GLOBALS['current_blog']) || empty($GLOBALS['current_blog']->registered)) {
        return;
    }

    $broker = $this->get_broker();
    // ... unchanged ...
}
```

**Behavior:**
- When `$current_blog` is fully populated → behavior unchanged (guard does not trigger).
- When `$current_blog->registered` is empty → method returns silently. The bearer-to-cookie conversion is re-attempted on the next request once WP has finished hydrating the blog. No SSO state is lost, no transient is touched.

**Why `empty()` and not `isset()`:** `isset($GLOBALS['current_blog']->registered)` returns true even when the property exists with an empty string (which is what WP sets it to during the race). `empty()` catches null, empty string, and missing-object cases in one check.

---

### Patch 2 — `calculate_secret_from_date()`

**File:** `inc/sso/class-sso.php`
**Method:** `public function calculate_secret_from_date($date)`
**Inserted:** 21 lines (comment block + 4-line fallback) at the top of the method body, before the `DateTimeZone` instantiation.

```php
public function calculate_secret_from_date($date) {

    /*
     * Fall back to the main site registration date when $date is empty.
     *
     * This guards against the same multisite bootstrap race that
     * `convert_bearer_into_auth_cookies()` skips: a caller can still
     * reach this method with an empty $date (for example, custom code
     * that builds an SSO secret from `$current_blog->registered` before
     * sunrise.php finishes populating it). Throwing SSO_Exception here
     * breaks the parent request, which on iframed admin pages renders
     * the whole panel blank.
     *
     * Using the main site registration date as a fallback keeps the
     * secret stable across requests for the same network and avoids
     * cascading failures during early boot.
     */
    if (empty($date)) {
        $main_site = function_exists('get_site') ? get_site(function_exists('get_main_site_id') ? get_main_site_id() : 1) : null;
        $date      = ($main_site && ! empty($main_site->registered)) ? $main_site->registered : '2024-01-01 00:00:00';
    }

    $tz = new \DateTimeZone('GMT');
    // ... unchanged ...
}
```

**Why a defense-in-depth second guard:**
- Patch 1 covers the only known production path that triggers the bug (the `init` callback).
- Patch 2 protects every other caller of `calculate_secret_from_date()` — third-party add-ons, custom integrations, and any future code path that might build secrets from `$current_blog->registered` outside the SSO bootstrap. Throwing in this method renders the whole parent request as a 500.

**Determinism:** the fallback uses `get_main_site_id()` → `get_site()->registered`, which is the canonical, stable network-scoped value. Two requests that hit the fallback in the same network will produce the same secret. Two requests that hit the fallback on different networks produce different secrets (correct isolation).

The literal `'2024-01-01 00:00:00'` is only reached if `get_site()` itself returns null (extremely degraded state — would mean WP core is half-loaded). Included only to avoid throwing inside a guard whose entire purpose is to prevent throwing.

**`function_exists()` guards** keep the fallback safe to call from contexts where `wp-includes/ms-load.php` may not yet be loaded.

---

### Edge cases verified

| Scenario | Behavior with patch |
|---|---|
| Healthy request, `$current_blog->registered` populated | Unchanged — guards do not fire |
| `$GLOBALS['current_blog']` is `null` | Returns early in `convert_bearer_into_auth_cookies()` |
| `$GLOBALS['current_blog']` is object, `registered = ''` | Returns early; if `calculate_secret_from_date()` is invoked directly with `''`, falls back to main site date |
| `$GLOBALS['current_blog']` is object, `registered = null` | Same as above (`empty()` covers both) |
| Main site has empty `registered` (broken install) | Falls back to literal `'2024-01-01 00:00:00'` instead of throwing |
| Subsite is suspended / deleted | No change — the patches do not interact with site status |
| User is not logged in | No change — the inner `is_user_logged_in() && $broker && $broker->isAttached()` block is reached only after the guard, same as before |

---

### Suggested unit tests

```php
// tests/SSO/SSOTest.php

public function test_calculate_secret_from_date_returns_main_site_fallback_for_empty_date(): void {
    $sso  = SSO::get_instance();
    $main = get_site(get_main_site_id());

    $expected = wp_hash((int) (new \DateTime($main->registered, new \DateTimeZone('GMT')))->format('mdisY'));
    $actual   = $sso->calculate_secret_from_date('');

    $this->assertSame($expected, $actual, 'Empty date should fall back to the main site registered date.');
}

public function test_calculate_secret_from_date_unchanged_for_valid_date(): void {
    $sso = SSO::get_instance();

    $a = $sso->calculate_secret_from_date('2024-11-06 15:24:32');
    $b = $sso->calculate_secret_from_date('2024-11-06 15:24:32');

    $this->assertSame($a, $b, 'Same date must produce same secret.');
    $this->assertNotSame($a, $sso->calculate_secret_from_date('2024-11-07 15:24:32'));
}

public function test_convert_bearer_skips_when_current_blog_not_populated(): void {
    global $current_blog;
    $original = $current_blog;

    $current_blog            = new \stdClass();
    $current_blog->registered = '';

    $sso = SSO::get_instance();
    $this->assertNull($sso->convert_bearer_into_auth_cookies(), 'Must return void without throwing when current_blog is half-populated.');

    $current_blog = $original;
}
```

---

### Reproduction on staging (see previous comment for the WP-CLI script)

- **UM 2.11.0 official (no patch):** `SSO_Exception :: La creación del secreto SSO falló` when `$current_blog->registered = ''`.
- **UM 2.11.0 + this PR:** returns deterministic secret `cc95d360e9559785ec2d094e951e54a2`, no exception.

Staging is kpstage.com, full clone of kursopro.com production: 200+ subsites, several mapped domains (test ran against `javieraacademy.com`, blog_id 218), PHP 8.3, WP 6.9, LiteSpeed + Cloudflare.

---

### Why this matters for 2.11.x

Networks that combine UM SSO with **any iframed admin** (WPFA / WP Frontend Admin, frontend admin panels, custom dashboards on mapped domains) will hit blank panel pages immediately after updating 2.10.1 → 2.11.0 on the first cache-miss request to a mapped-domain subsite. The bug is invisible in dev (no Cloudflare, no LiteSpeed, no SaaS-scale cache invalidation) but surfaces predictably in production whenever the page cache cycles.

The patch is **41 lines, single file, zero behavior change on healthy requests, validated in production for 5+ days**. It is the minimum surface needed to make 2.11.0 drop-in safe for SaaS-style multisite installs.

Let me know if you want this rebased on any specific branch, or if you prefer the unit tests in a follow-up PR.

superdav42:
## Stuck-merge detector: PR has been merge-eligible but unmerged past the threshold

The pulse merge pass has classified PR #1185 as `STUCK_OTHER` and it has been sitting unmerged longer than `AIDEVOPS_MERGE_STUCK_AGE_MINUTES` (currently 240m). The deterministic merge gates are evaluated every cycle (~120s) and this PR has consistently failed them.

### Failing checks on PR #1185

- (no FAILURE entries in rollup; check rollup manually)

### Worker guidance for the next attempt

1. Read PR #1185 body + the latest check run logs:
   ```bash
   gh pr view 1185 --repo Ultimate-Multisite/ultimate-multisite --json statusCheckRollup
   gh pr checks 1185 --repo Ultimate-Multisite/ultimate-multisite
   ```
2. If the failing checks are environment/Setup-step (Format, Lint, Typecheck all FAIL at the same step), the canonical default branch likely has a broken lockfile or a CI infra change — fix at the base, not on this PR. Look for a sibling outage meta-issue in this repo (filed by the same detector) before forking off here.
3. If the failures are PR-specific (e.g. a Typecheck error introduced by this PR's code), rebase onto the latest default branch and address the diagnosed errors. Use `full-loop-helper.sh start` from the linked PR's worktree.
4. If the linked issue body lacks the worker-ready file paths and verification commands required by t1900, post a comment naming the missing context before dispatching another worker — the next attempt will burn tokens on exploration otherwise.

### Why you're seeing this

Every pulse cycle (~120s) the deterministic merge pass re-evaluates open PRs. PRs that pass APPROVED + MERGEABLE but fail required checks have historically been re-evaluated silently every cycle until a human noticed. The stuck-merge detector (t3193) surfaces them after `AIDEVOPS_MERGE_STUCK_AGE_MINUTES` minutes idle. This comment is posted exactly once per linked issue — repeated stuck cycles will NOT spam the thread. If the PR merges and the issue is reopened later with a fresh stuck PR, the marker will allow a second comment.

<sub>Posted automatically by `pulse-merge-stuck.sh` (t3193 / GH#21895). Threshold env: `AIDEVOPS_MERGE_STUCK_AGE_MINUTES=240`.</sub>


---
[aidevops.sh](https://aidevops.sh) v3.15.34 automated scan.

PR #1198, opened 2026-05-12
test(e2e): regression guards for #1185 (SSO bootstrap race) and #1169 (subsite password reset)

kenedytorcatt:
## Why this PR exists

A senior contributor on our side recently shipped two fixes that affected ~300+ subsites in our production network (kursopro.com), and we'd like to make sure they can't silently regress in a future refactor:

- **PR #1185** — `inc/sso/class-sso.php` early-return + deterministic-secret fallback (merged 2.11.1).
- **PR #1169** — `inc/checkout/class-checkout-pages.php` subsite password reset URL rewrite (merged 2.10.2).

This PR adds two Cypress specs + matching WP-CLI fixtures that lock in the behaviour of those fixes. They sit alongside your existing `060-sso-cross-domain.spec.js`, `065-sso-redirect-loop.spec.js`, and `050-password-strength-enforcement.spec.js` and follow the same patterns (`cy.wpCliFile`, JSON output from fixture, `cy.loginByApi`, baseUrl `localhost:8889`).

### `066-sso-bootstrap-race.spec.js` — guards #1185

- `calculate_secret_from_date('')` does **not** throw and returns a real hash.
- Two consecutive calls with empty input return the **same** hash (deterministic fallback — important so SSO state stays consistent across requests during the empty-\`registered\` window).
- `convert_bearer_into_auth_cookies()` does **not** throw when \`\$current_blog\` exists with an empty \`registered\` property.

Driven entirely from PHP via `setup-sso-bootstrap-race.php`. The race is at the PHP layer (between `sunrise.php` and `ms-settings.php`), so no browser interaction is required.

### `011-password-reset-subsite-domain.spec.js` — guards #1169

- The URL produced by the `retrieve_password_message` filter on a subsite uses the **subsite** host — not the main-site \`/wp-login.php\` host (the symptom of the original bug that confused `zuletadia` on `yariglam.cl`).
- The reset query args \`action=rp / key / login / wp_lang\` are preserved so WooCommerce my-account, BuddyPress, custom themes, and the default wp-login fallback can still complete the reset.
- The new `wu_subsite_password_reset_url` filter is reachable for integration overrides.

Driven via `setup-password-reset-subsite.php` which creates a test subsite, switches into it, and applies the filter chain directly with a synthetic raw message — no SMTP / Mailpit / cron dependency.

## How they fit your existing suite

| File | Type | Notes |
|---|---|---|
| `tests/e2e/cypress/integration/066-sso-bootstrap-race.spec.js` | new | Numbered after your `060/065` SSO specs |
| `tests/e2e/cypress/integration/011-password-reset-subsite-domain.spec.js` | new | Numbered next to the `010-manual-checkout-flow.spec.js` group |
| `tests/e2e/cypress/fixtures/setup-sso-bootstrap-race.php` | new | Same pattern as your `setup-sso-test.php` |
| `tests/e2e/cypress/fixtures/setup-password-reset-subsite.php` | new | Same pattern as your `setup-sso-test.php` |

Both specs are deterministic, single-process, and avoid any heavy setup (no Stripe, no PayPal, no Mailpit).

## Honest note — please validate in your CI before merging

I authored these on Windows without Docker locally, so **I was not able to run `npm run cy:run:test` against your `wp-env` setup**. I verified:

- PHP `-l` syntax check on both fixtures.
- `node --check` syntax check on both specs.
- That the SSO class is a singleton (`\WP_Ultimo\SSO\SSO::get_instance()`) by reading `inc/sso/class-sso.php`.
- That `replace_reset_password_link()` and the `wu_subsite_password_reset_url` filter live where the PR diffs say they do.
- That fixture conventions match your `setup-sso-test.php` (JSON output, `wp_json_encode`, container path via `cy.wpCliFile`).

If anything fails in your CI I'm happy to iterate — feel free to push directly to the branch or comment with the failure output.

## Production context

Bug context behind each spec, in case it helps reviewing the assertions:

- **#1185 SSO bootstrap race**: surfaced as blank admin panels inside iframes on mapped-domain subsites after a cache invalidation. Masked by full-page cache on healthy requests, reappeared every time the cache was cold. Affected 300+ subsites until 2.11.1.
- **#1169 password reset on subsites**: a customer of a subsite owner (mapped `.cl` TLD) requested a password reset and received an email pointing at the **main network**'s `wp-login.php`. She didn't recognise the domain, treated the email as phishing, and complained to the subsite owner that the site had been hacked. Reproduced on 8/8 subsites in our staging network before patch.

Thanks David — and thank you for shipping both fixes so quickly. These specs are our way of helping keep them locked in.

### DIFF
--- a/tests/e2e/cypress/fixtures/setup-password-reset-subsite.php
+++ b/tests/e2e/cypress/fixtures/setup-password-reset-subsite.php
@@ -1,0 +1,175 @@
+<?php
+/**
+ * Reproduce the subsite password reset URL bug fixed in PR #1169 / issue #1168.
+ *
+ * The bug:
+ *   `replace_reset_password_link()` had an early return on subsites and the
+ *   `retrieve_password_message` filter registration was gated by
+ *   `is_main_site()`. End users who triggered "lost password" on a subsite
+ *   (especially mapped custom domains) received an email pointing to
+ *   `https://main-site.example/wp-login.php` — a domain they don't recognise.
+ *
+ * What this fixture does:
+ *   1. Ensures custom login pages are enabled (the affected setting).
+ *   2. Creates a test subsite at `/pwreset-subsite/` if missing.
+ *   3. Creates a test user owned by that subsite.
+ *   4. Switches to the subsite context and invokes UM's filter directly
+ *      with a synthetic raw message containing a main-site wp-login.php URL.
+ *   5. Extracts the URL from the filtered message and returns its host /
+ *      path / query so the Cypress spec can assert against them.
+ *   6. Also asserts the new `wu_subsite_password_reset_url` filter (added
+ *      by #1169) is reachable — by hooking it transiently and confirming
+ *      it ran.
+ *
+ * Returns JSON consumed by the Cypress spec:
+ *   {
+ *     "site_id":          <int>,
+ *     "subsite_host":     "<host expected to appear in the reset URL>",
+ *     "main_host":        "<main site host>",
+ *     "filter_ran":       true,                  // wu_subsite_password_reset_url
+ *     "rewritten_url":    "<final URL pulled from the message>",
+ *     "rewritten_host":   "<host of the rewritten URL>",
+ *     "rewritten_query":  { action, key, login, wp_lang },
+ *     "raw_url":          "<the URL we asked UM to rewrite>"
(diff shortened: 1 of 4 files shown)


superdav42 (approved):

PR #1267, opened 2026-05-23
fix(checkout): honour is_publishing_stale in publish_pending_site() to unblock stuck pending sites

superdav42:
## Summary

`publish_pending_site()` in `inc/models/class-membership.php` bailed
silently when `is_publishing=true`, even after the publishing process
that set the flag had died. The result: the "Creating your site"
overlay hung on the customer's screen until the next Action Scheduler
retry — which often never fires, because the duplicate caller never
reschedules.

This change wires `Site::is_publishing_stale()` (added in 2.5.3 to
detect this exact "flag set but the process that set it is dead"
condition, default 5-minute window) into `publish_pending_site()` so
a second caller can finish the job the first caller never completed.
The AJAX poller at `class-membership-manager.php:214` already consumed
`is_publishing_stale()`; this commit closes the symmetry on the
publish path.

A fresh in-flight flag still short-circuits, so the duplicate-call
short-circuit semantics are unchanged for the non-stale case.

## Trigger

A real-world reproduction observed in production logs: a duplicate
$0-trial WooCommerce order arrives ~2.5 minutes after the first order
fails mid-publish (PHP process killed by `max_execution_time` / OOM /
fatal error). Order #1 set `is_publishing=true` and died; order #2's
`publish_pending_site()` call saw the flag and returned without doing
anything. The site stayed at `type=site_template` and the customer's
overlay never completed.

The duplicate-order root cause itself lives in the
`ultimate-multisite-woocommerce` addon; this is the core defence-in-depth
fix so the symptom self-heals after the 5-minute stale window.

## Patch

```php
if ($is_publishing) {
    $is_stale = method_exists($pending_site, 'is_publishing_stale')
        ? $pending_site->is_publishing_stale()
        : false;

    if ( ! $is_stale) {
        return true;
    }

    wu_log_add(/* … stale-recovery log line … */);
}
```

`method_exists()` guards the call so any pre-2.5.3 serialized `Site`
objects deserialized from membership meta still behave like the old
code path. The recovery emits a `wu_log_add()` line on the membership
log so support can correlate stuck-overlay reports.

## Regression coverage

`tests/WP_Ultimo/Models/Membership_Test.php`

- `test_publish_pending_site_bails_when_flag_is_fresh` — a fresh flag
  must still short-circuit (guards against accidentally rewriting the
  check into "always publish").
- `test_publish_pending_site_proceeds_when_flag_is_stale` — a stale
  flag must fall through and actually publish the blog. This is the
  regression guard for the BUG fix; reverting just the inc/ patch and
  re-running this test fails as expected.

Both tests use `wu_create_customer` / `wu_create_product` /
`wu_create_membership` (BerlinDB-persisting helpers); the class-level
`setUp()` uses raw Model constructors which can leave the in-memory ID
at 0 in some environments, silently no-op'ing the meta calls these
tests depend on.

## Verification

- PHPCS on changed inc/ file: clean
- PHPCS on new test block: clean (lines 1764+; pre-existing warnings
  elsewhere in the test file are not touched)
- PHPStan level 0: `[OK] No errors`
- New tests: `OK (2 tests, 14 assertions)`
- Full `Membership_Test` class (118 tests): green
- Deterministic WP-CLI reproduction script confirmed the bug
  pre-patch and confirmed the fix post-patch on the dev install

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of pending site publishing when a publishing operation appears to be running but has stalled. The system now detects stale publishing states and completes the process.

* **Tests**
  * Added test coverage for pending site publishing recovery scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Ultimate-Multisite/ultimate-multisite/pull/1267?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/inc/models/class-membership.php
+++ b/inc/models/class-membership.php
@@ -2178,7 +2178,41 @@ class Membership extends Base_Model implements Limitable, Billable, Notable {
 		$is_publishing = $pending_site->is_publishing();
 
 		if ($is_publishing) {
-			return true;
+			/*
+			 * If is_publishing has been true for longer than the stale
+			 * timeout (default 5 minutes), the PHP process that set the
+			 * flag is presumed dead (OOM, max_execution_time, fatal,
+			 * server restart). The poller resets the flag in that case,
+			 * but if a competing call (e.g. a duplicate checkout order)
+			 * lands here while the flag is still set we would otherwise
+			 * silently bail and the "Creating your site" overlay would
+			 * hang until the next Action Scheduler retry — which often
+			 * never fires because the duplicate caller never reschedules.
+			 *
+			 * Detect the stale flag here and fall through to the publish
+			 * path so the second caller can finish the job the first
+			 * caller never completed. Site::is_publishing_stale() was
+			 * added in 2.5.3; method_exists() keeps this safe for any
+			 * pre-2.5.3 serialized Site objects deserialized from meta.
+			 *
+			 * @since 2.5.4
+			 */
+			$is_stale = method_exists($pending_site, 'is_publishing_stale')
+				? $pending_site->is_publishing_stale()
+				: false;
+
+			if ( ! $is_stale) {
+				return true;
+			}
+
+			wu_log_add(
+				"membership-{$this->get_id()}",
+				sprintf(
+					// translators: %d: membership ID.
+					__('Detected stale is_publishing flag on pending site for membership %d during publish_pending_site(); the previous publish attempt appears to have been killed before completing. Proceeding with retry.', 'ultimate-multisite'),
+					$this->get_id()
+				)
+			);
 		}
 
 		$pending_site->set_publishing(true);
--- a/tests/WP_Ultimo/Models/Membership_Test.php
+++ b/tests/WP_Ultimo/Models/Membership_Test.php
@@ -1761,4 +1761,168 @@ class Membership_Test extends \WP_UnitTestCase {
 		$this->assertInstanceOf(Membership::class, $captured);
 		$this->assertSame($m->get_id(), $captured->get_id());
 	}
+
+	// ---------------------------------------------------------------
+	// publish_pending_site() — stale is_publishing flag handling
+	//
+	// Regression coverage for the "Creating your site" overlay hang
+	// caused by a duplicate caller bailing on a stuck is_publishing
+	// flag. See inc/models/class-membership.php::publish_pending_site().
+	// ---------------------------------------------------------------
+
+	/**
+	 * Build a persisted membership via the wu_create_* helpers (which
+	 * write through BerlinDB and return a model with a real ID) so that
+	 * update_meta/get_meta operate against a real DB row.
+	 *
+	 * The class-level $this->membership built in setUp() uses raw Model
+	 * constructors with skip_validation; in some environments the save()
+	 * path leaves the in-memory ID at 0, which makes meta calls silently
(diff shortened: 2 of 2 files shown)


superdav42:
## Summary
`publish_pending_site()` in `inc/models/class-membership.php` bailed
silently when `is_publishing=true`, even after the publishing process
that set the flag had died. The result: the "Creating your site"
overlay hung on the customer's screen until the next Action Scheduler
retry — which often never fires, because the duplicate caller never
reschedules.
This change wires `Site::is_publishing_stale()` (added in 2.5.3 to
detect this exact "flag set but the process that set it is dead"
condition, default 5-minute window) into `publish_pending_site()` so
a second caller can finish the job the first caller never completed.
The AJAX poller at `class-membership-manager.php:214` already consumed
`is_publishing_stale()`; this commit closes the symmetry on the
publish path.
A fresh in-flight flag still short-circuits, so the duplicate-call
short-circuit semantics are unchanged for the non-stale case.
## Trigger
A real-world reproduction observed in production logs: a duplicate
$0-trial WooCommerce order arrives ~2.5 minutes after the first order
fails mid-publish (PHP process killed by `max_execution_time` / OOM /
fatal error). Order #1 set `is_publishing=true` and died; order #2's
`publish_pending_site()` call saw the flag and returned without doing
anything. The site stayed at `type=site_template` and the customer's
overlay never completed.
The duplicate-order root cause itself lives in the
`ultimate-multisite-woocommerce` addon; this is the core defence-in-depth
fix so the symptom self-heals after the 5-minute stale window.
## Patch
```php
if ($is_publishing) {
    $is_stale = method_exists($pending_site, 'is_publishing_stale')
        ? $pending_site->is_publishing_stale()
        : false;
    if ( ! $is_stale) {
        return true;
    }
    wu_log_add(/* … stale-recovery log line … */);
}
```
`method_exists()` guards the call so any pre-2.5.3 serialized `Site`
objects deserialized from membership meta still behave like the old
code path. The recovery emits a `wu_log_add()` line on the membership
log so support can correlate stuck-overlay reports.
## Regression coverage
`tests/WP_Ultimo/Models/Membership_Test.php`
- `test_publish_pending_site_bails_when_flag_is_fresh` — a fresh flag
  must still short-circuit (guards against accidentally rewriting the
  check into "always publish").
- `test_publish_pending_site_proceeds_when_flag_is_stale` — a stale
  flag must fall through and actually publish the blog. This is the
  regression guard for the BUG fix; reverting just the inc/ patch and
  re-running this test fails as expected.
Both tests use `wu_create_customer` / `wu_create_product` /
`wu_create_membership` (BerlinDB-persisting helpers); the class-level
`setUp()` uses raw Model constructors which can leave the in-memory ID
at 0 in some environments, silently no-op'ing the meta calls these
tests depend on.
## Verification
- PHPCS on changed inc/ file: clean
- PHPCS on new test block: clean (lines 1764+; pre-existing warnings
  elsewhere in the test file are not touched)
- PHPStan level 0: `[OK] No errors`
- New tests: `OK (2 tests, 14 assertions)`
- Full `Membership_Test` class (118 tests): green
- Deterministic WP-CLI reproduction script confirmed the bug
  pre-patch and confirmed the fix post-patch on the dev install

---
Merged via PR #1267 to main.

---
[aidevops.sh](https://aidevops.sh) v3.17.28 spent 32s on this as a headless bash routine.

PR #1203, opened 2026-05-13
GH#1199: fix: cap SSO redirects on mapped domains

superdav42:
## Summary

Added a short-window SSO redirect-attempt counter so mapped-domain admin requests fall back to the normal login form instead of looping when broker attachment cannot complete. Added SSO unit coverage for threshold and window-reset behaviour.

## Files Changed

inc/sso/class-sso.php,tests/WP_Ultimo/SSO/SSO_Test.php

## Runtime Testing

- **Risk level:** Low (agent prompts / infrastructure scripts)
- **Verification:** WP_TESTS_DIR=/tmp/wordpress-tests-lib vendor/bin/phpunit --filter SSO_Test; vendor/bin/phpcs inc/sso/class-sso.php tests/WP_Ultimo/SSO/SSO_Test.php (warnings only: pre-existing file_get_contents/base64 notices); vendor/bin/phpstan analyse inc/sso/class-sso.php tests/WP_Ultimo/SSO/SSO_Test.php

Resolves #1199

---
[aidevops.sh](https://aidevops.sh) v3.15.38 plugin for [OpenCode](https://opencode.ai) v1.14.48 with gpt-5.5 spent 5m and 122,378 tokens on this with the user in an interactive session.

## Release Notes

* **Bug Fixes**
  * Added redirect-loop detection for SSO authentication to prevent infinite redirects when using mapped domains.
  * Configured redirect-loop detection threshold and time window via filters for customization.

* **Tests**
  * Added test coverage for SSO redirect-loop throttling behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Ultimate-Multisite/ultimate-multisite/pull/1203)

PR #1084, opened 2026-05-04
fix: cookie-less cross-domain SSO login flow

superdav42:
## Summary

Enables SSO login from subsite to main site without requiring 3rd party cookies.

### Problem

When visiting a subsite (e.g., vidanuevanaz.org) wp-admin, users were redirected to the main site (mygratis.site) login page. After logging in, the redirect back to the subsite didn't work without 3rd party cookies enabled.

### Solution

1. **Token-based authentication**: After login on main site, generate a time-limited `wu_sso_token` and add it to the return URL when redirecting to a different domain.

2. **Subsite token validation**: The multi-tenancy plugin's `handle_sso_login()` now accepts simpler token format from the main site's SSO handler.

3. **Already-logged-in handling**: When user is already logged in on main site and visits login page with SSO params, redirect directly to subsite with token instead of showing "already logged in" message.

### Changes

- `class-sso.php`: Add `generate_sso_token()`, `handle_login_redirect()` extracts return_url from query params, `handle_already_logged_in_on_login_page()` handles logged-in users on login page
- `class-remote-sso-handler.php` (multi-tenancy): Add `validate_sso_token_from_main_site()` to accept simpler token format

### Testing

1. Visit `vidanuevanaz.org/wp-admin/` (unauthenticated)
2. Should redirect to `mygratis.site/login/?sso=login&return_url=https://vidanuevanaz.org`
3. Log in with credentials
4. Should redirect back to `vidanuevanaz.org` - logged in

Also test already-logged-in flow:
1. Already logged in on main site
2. Visit subsite wp-admin
3. Should redirect directly to subsite dashboard

---
[aidevops.sh](https://aidevops.sh) v3.14.51 plugin for [OpenCode](https://opencode.ai) v1.14.33 with claude-sonnet-4-6 spent 15h 33m on this as a headless worker.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced SSO authentication flow to properly recognize and handle return URL parameters during login.
  * Improved redirect destination extraction from URL query strings to ensure users are reliably directed to their intended location after successful authentication.
  * Refined redirect logic conditions to support more flexible parameter configurations for authentication flows.

PR #1297, opened 2026-05-27
fix(sso): restore front-end auto-SSO via main-site magic-link redirect

superdav42:
## Summary

Visiting a subsite front-end page while logged in to the main site was
not re-authenticating the visitor on the subsite — even in browsers
that permit third-party cookies. Two interacting regressions blocked
the flow:

1. **#1185** turned the JSONP unattached branch in `handle_broker()` into
   an immediate `{code: 0, message: "Broker not attached"}` response,
   which calls `sso_denied()` in `sso.js` and sets `wu_sso_denied` for
   5 minutes on the very first subsite page-load. After that there is
   no path that ever surfaces the main-site session to the broker.

2. The redirect branch still called `$broker->getAttachUrl()` which
   targets `/sso-grant`. After the cookie-less rework in #1084 the
   dispatcher in `handle_requests()` unconditionally appends `_grant`
   on main, so `/sso-grant` produces `wu_sso_handle_sso_grant_grant`
   — an action with no listener — and the request silently 404s on
   main, leaving the user logged out on the subsite.

## What this changes

Both branches now redirect to the main-site `/sso` endpoint that
`handle_server()` already supports under the cookie-less rework.
`handle_server()` checks the **first-party** main-site auth cookie (so
third-party-cookie posture is irrelevant), issues an HMAC-signed
`wu_sso_token` via `add_cookie_less_sso_token()`, and 302s back to the
subsite where `handle_cookie_less_sso_token()` consumes the token at
`init` priority 4 and sets the broker-side auth cookie.

Resulting request chain (verified in browser):

```
GET <broker>/sso?_jsonp=1                              -> 200 JSONP (must-redirect)
GET <broker>/sso?return_url=<original_subsite_url>     -> 302
GET <main>/sso?sso=login
     &return_url=<original>&redirect_to=<original>     -> 302
GET <broker>/?wu_sso_token=<hmac>&redirect_to=<orig>   -> 302
GET <broker>/                                          -> 200, logged in
```

### JSONP branch

The JSONP `<script>` tag cannot follow a 302 to HTML, so instead of
silently failing the broker returns `verify: 'must-redirect'`. The
already-present branch in `assets/js/sso.js` (lines 93-95) handles this
by performing a top-level `window.location.replace()` to
`o.server_url + '?return_url=<encoded current>'`, which re-enters
`handle_broker()` via the non-JSONP path on the same request and
redirects to `$main_sso_url`.

### Return-URL handling

- The forwarded `return_url` is taken from the explicit `?return_url=`
  query param when present (set by `sso.js` on the must-redirect
  navigation) and falls back to `get_current_url()` only on the direct
  first hit.
- The `/sso` path itself is rejected as a return target — without
  this guard the value `/sso?return_url=…` would feed back into
  `return_url` on each iteration and the URL would grow on every hop,
  producing an `ERR_TOO_MANY_REDIRECTS`.
- `redirect_to` is passed explicitly so `get_sso_redirect_to()` does
  not append `/wp/wp-admin/` (its fallback when only a cross-domain
  `return_url` is supplied) and therefore does not send front-end
  visitors to the subsite admin.

## Tests / verification

- `php -l` clean on the modified file.
- Verified end-to-end in a real browser against a Bedrock multisite
  with mapped subsite domains: a main-site-logged-in visitor lands
  logged-in on the subsite front-end page they originally requested.
- Re-running the same flow without clearing `wu_sso_denied` is
  unaffected by the previous broken behaviour because the new code
  path no longer sets that cookie on its own.

## Out of scope

Two related issues observed while diagnosing this, left for follow-up:

1. **`/sso-grant` routing is dead.** The dispatcher in
   `handle_requests()` rewrites the action to `sso_grant` and then
   appends `_grant`, so the only URL that reaches `handle_server()` is
   `/sso` (not `/sso-grant`). The `SSO_Broker::getAttachUrl()` method
   still points at `/sso-grant`. After this PR no caller uses that URL,
   but it should either be removed or the dispatcher should normalise
   so the two URLs do the same thing.
2. **`should_skip_redirect_due_to_loop()` from #1203** increments
   `wu_sso_redirect_attempts` on every logged-out `handle_auth_redirect`
   call regardless of whether the previous request was itself part of
   an SSO loop. It can burn the counter on benign repeat visits to
   wp-admin from a logged-out user. Consider scoping it to requests
   whose `Referer` is the main-site `/sso*` endpoint.

---
[aidevops.sh](https://aidevops.sh) v3.20.0 plugin for [OpenCode](https://opencode.ai) v1.15.11 with claude-opus-4-7 spent 1h 2m and 81,144 tokens on this with the user in an interactive session.

### DIFF
--- a/assets/js/sso.js
+++ b/assets/js/sso.js
@@ -24,6 +24,7 @@
 	const checkout_form = document.querySelector(
 		'#wrapper-field-checkout, .wu-checkout, .wu-checkout-form'
 	);
+	const checkout_url = /\/register\/?$/.test(window.location.pathname);
 
 	document.head.insertAdjacentHTML('beforeend', `
     <style>
@@ -63,7 +64,7 @@
     </style>
   `);
 
-	if (! o.is_user_logged_in && ! denied && ! checkout_form) {
+	if (! o.is_user_logged_in && ! denied && ! checkout_form && ! checkout_url) {
 
 		const s = document.getElementsByTagName('script')[ 0 ];
 
--- a/assets/js/sso.js
+++ b/assets/js/sso.js
@@ -21,6 +21,10 @@
 
 	const denied = wu_read_cookie('wu_sso_denied');
 
+	const checkout_form = document.querySelector(
+		'#wrapper-field-checkout, .wu-checkout, .wu-checkout-form'
+	);
+
 	document.head.insertAdjacentHTML('beforeend', `
     <style>
       @keyframes fade_in {
@@ -59,7 +63,7 @@
     </style>
   `);
 
-	if (! o.is_user_logged_in && ! denied) {
+	if (! o.is_user_logged_in && ! denied && ! checkout_form) {
 
 		const s = document.getElementsByTagName('script')[ 0 ];
 
--- a/inc/sso/class-sso.php
+++ b/inc/sso/class-sso.php
@@ -945,13 +945,87 @@ class SSO {
 
 		// Attach through redirect if the client isn't attached yet.
 		if ( ! $broker->isAttached()) {
+			$sso_path = $this->get_url_path();
+
 			/*
-			 * For JSONP requests (initiated by a <script> tag), we must NOT
-			 * redirect — the browser follows 302s transparently for script
-			 * tags, but the final response from the server will be a redirect
-			 * (not JavaScript), so the wu.sso() callback never fires.
-			 * Instead, return a JSONP error so the JS can handle it gracefully
-			 * and set the wu_sso_denied cookie to prevent further attempts.
+			 * Determine the page we ultimately want the user to land on
+			 * after the magic-link round-trip.
+			 *
+			 * The JSONP must-redirect fallback navigates the top-level
+			 * browser to `<broker>/sso?return_url=<original_page>`, so on
+			 * that second hit `get_current_url()` is the /sso URL itself
+			 * — using it would feed `/sso?...` back into return_url and
+			 * the main site would echo a longer `/sso?return_url=/sso?...`
+			 * URL on each iteration, producing a `ERR_TOO_MANY_REDIRECTS`
+			 * (the URL grows on every hop, never converges).
+			 *
+			 * Prefer the explicit `return_url` query param when present
+			 * (set by sso.js on the must-redirect branch); only fall back
+			 * to `get_current_url()` on a direct first hit. Guard the
+			 * value against pointing at the /sso path itself so a
+			 * stray/legacy URL can't restart the same loop.
+			 */
+			$requested_return = (string) $this->input('return_url', '');
+
+			if ( '' === $requested_return ) {
+				$return_url = $this->get_current_url();
+			} else {
+				$return_url = $requested_return;
+			}
+
+			$return_path = (string) wp_parse_url($return_url, PHP_URL_PATH);
+
+			if ( '' === $return_path || preg_match("#/{$sso_path}(-grant)?/?$#", $return_path) ) {
+				// Refuses to round-trip the /sso endpoint itself; fall back to subsite home.
+				$return_url = home_url('/');
+			}
+
+			/*
+			 * Front-end auto-SSO target: send the user to the main site's
+			 * /sso endpoint with a return_url + redirect_to. handle_server()
+			 * on the main site checks the main-site auth cookie (first-party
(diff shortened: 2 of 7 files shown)


superdav42:
Follow-up issues filed for the two items in the "Out of scope" section above:

- #1298 — `/sso-grant` URL produced by `getAttachUrl()` is no longer routed
- #1299 — `should_skip_redirect_due_to_loop()` over-counts on non-SSO requests


---
[aidevops.sh](https://aidevops.sh) v3.20.0 plugin for [OpenCode](https://opencode.ai) v1.15.11 with claude-opus-4-7 spent 1h 11m and 87,936 tokens on this with the user in an interactive session.

superdav42:
CLAIM_RELEASED reason=worker_failed runner=superdav42 ts=2026-05-28T04:39:04Z aidevops_version=3.20.4 opencode_version=1.15.11

superdav42:
## MERGE_SUMMARY

Status: BLOCKED — Cypress still fails on the target SSO spec after pushed fixes.

Changes pushed to `fix/sso-frontend-magic-link-flow`:
- `tests/e2e/cypress/fixtures/setup-sso-test.php`: attempts to make the wp-env SSO subsite resolve on `127.0.0.1:8889` by seeding a native test host and mapped-domain rows.
- `tests/e2e/cypress/integration/060-sso-cross-domain.spec.js`: aligns assertions with the mapped-host root-domain setup.
- `sunrise.php`: adds a dynamic fallback to locate `inc/class-sunrise.php` when wp-env maps the plugin under a non-standard directory.

Verification:
- `php -l tests/e2e/cypress/fixtures/setup-sso-test.php` — pass.
- `vendor/bin/phpcs tests/e2e/cypress/fixtures/setup-sso-test.php` — pass.
- `php -l sunrise.php` — pass.
- `vendor/bin/phpcs sunrise.php` — pass.
- GitHub checks on latest pushed commit `bbe87789`: PHP Lint, PHP 8.2/8.3/8.4/8.5, Code Quality Checks, Build Plugin for Testing — pass.
- GitHub E2E run `26584302903`: `cypress (8.2, chrome)` still fails in `tests/e2e/cypress/integration/060-sso-cross-domain.spec.js`; the response for `http://127.0.0.1:8889/` remains `302` to `http://localhost:8889/` instead of resolving to the test subsite.

Current blocker:
- The CI wp-env request to `127.0.0.1:8889` is still handled as an unknown host and redirected to the main site before the SSO assertions can execute, so the target Cypress check is not green yet.

superdav42:
CLAIM_RELEASED reason=worker_complete runner=superdav42 ts=2026-05-28T15:38:53Z aidevops_version=3.20.4 opencode_version=1.15.11

superdav42:
Closing in favour of #1309, which carries the minimal subset of this PR that the post-#1301 / post-#1302 codebase still needs.

When this PR was opened, two regressions interacted:

1. The JSONP unattached branch returned `{code: 0, message: "Broker not attached"}`, which tripped `sso.js` into setting `wu_sso_denied` for 5 minutes on the very first subsite page-load.
2. `$broker->getAttachUrl()` produced `/sso-grant`, which the dispatcher then routed to the unlistened action `wu_sso_handle_sso_grant_grant`.

(2) has since been fixed independently:

- #1301 — `fix: route sso grant attach urls` — normalises `/sso-grant` so it reaches `wu_sso_handle_sso_grant`.
- #1302 — `fix(sso): scope redirect loop counter to SSO hops` — addresses the "out of scope" loop-counter follow-up flagged in this PR's description.

With those landed, the only remaining defect is the JSONP denial payload (1). #1309 fixes exactly that one line and adds source-code regression tests asserting `verify: must-redirect` is returned and that `sso.js` still handles it. The broader rewrite proposed here (redirecting through `/sso` instead of `/sso-grant`, plus the `sso.js` checkout-form skip, `sunrise.php` glob fallback, and wp-env Cypress fixture changes) is no longer required for this bug and was also blocking on a pre-existing wp-env environment issue in Cypress.

If the checkout-form skip and wp-env Cypress fixture rework are still wanted, they should each be opened as separate PRs scoped to their own bug.

superdav42:
CLAIM_RELEASED reason=process_exit runner=superdav42 ts=2026-06-02T22:43:21Z aidevops_version=3.20.10 opencode_version=1.15.13 exit=0 session_count=1

PR #1336, opened 2026-06-02
fix(checkout): harden step validation scoping

### LINKED ISSUE #1332: checkout: email/username/password required errors fire on template-selection step for guest users
**Bug:** Checkout validation errors for email/username/password fire on template selection step for guest users
**Plugin version:** v2.12.0
**PHP:** 8.x

## Steps to reproduce

1. Use a multi-step checkout form with account details on one step and template selection on the next
2. Visit the checkout page as a **guest** (not logged in)
3. Fill in email, username, and password on the account step and click **Next Step**
4. On the template selection step, select a template and click **Next Step**

## Expected behaviour

Advancing from the template selection step should validate only the fields on that step. Email, username, and password were already submitted on a prior step and saved to session - they should not be re-validated here.

## Actual behaviour

Every guest user sees the following errors on the template selection step:

- Email address is required.
- Username is required.
- Password is required.

New customer signups are completely blocked at this step.

## Root cause (identified in source)

In `assets/js/checkout.js`, `validate_client_side()` filters validation rules to the current step's fields via `wu_checkout.step_fields`. However, `validation_rules` still contains `required_without:user_id` for `email_address`, `username`, and `password`. For guests, `user_id` is always absent from the form - the hidden `<input name="user_id">` is only rendered in `views/checkout/form.php` when `is_user_logged_in()`. So the `required_without` condition is satisfied, and since the account fields are not in the DOM on the template step, their values are `''`, triggering all three errors.

```js
// validate_client_side() in assets/js/checkout.js
// step_fields['site'] = ['site_title', 'site_url', 'template_selection']
// email_address / username / password are correctly excluded from step rules...
// ...but required_without:user_id fires because user_id = '' for all guests
const form_values = Object.assign({}, form_data_obj, {
    user_id: form_data_obj.user_id || '',  // always '' for guests
});
```

This is the same root cause as #762 (which triggers on the plan selection step). The fix in PR #763 addresses both, but is not included in v2.12.0.

## Expected fix

PR #763 resolves this. In the interim, the `wu_checkout_js_validation_rules` filter can be used as a workaround:

```php
add_filter('wu_checkout_js_validation_rules', function($rules) {
    unset($rules['email_address'], $rules['username'], $rules['password']);
    return $rules;
});
```

## Notes

- Does **not** affect logged-in users - they have a rendered `user_id` hidden input, so `required_without` is never satisfied
- Server-side validation in `inc/checkout/class-checkout.php` is unaffected and correctly enforces account fields on the account step only - no security regression from the workaround
- Confirmed in v2.12.0 (the BCC fix release). The client-side validation layer that introduced this was not present in earlier versions

## Related

- #762 - same root cause, triggers on plan selection step instead
- PR #763 - fix pending merge

superdav42:
## Summary

- Hardens checkout client-side validation scoping so stale or missing `wu_checkout.step_fields` falls back to fields rendered on the current form step instead of validating every global rule.
- Adds validation-rule aliases to `step_fields` for checkout field types whose rendered/value key differs from the builder field id (`template_selection` → `template_id`, `pricing_table` → `products`).
- Extends the step-fields regression test with an account → template/site → payment flow to guard against revalidating guest email/username/password on the template step.

## Issue verification

- Verified PR #763 is already included in `v2.12.0` (`git tag --contains a3a6f7c...` returned `v2.12.0`), so the issue body's release-status note was incorrect.
- Browser Agent local reproduction with a guest multi-step checkout reached the later step with `errors: []`; the exact email/username/password client errors were not reproducible on current `v2.12.0` code.
- The same browser run showed the template step still used the raw `template_selection` field id in `step_fields`, while the JS validation rule key is `template_id`; this PR fixes that mismatch and adds a defensive fallback for stale step maps.

## Verification

- `vendor/bin/phpunit --filter Checkout_Step_Fields_Test` — passes (2 tests, 31 assertions; existing PHP 8.4 deprecation notices emitted by dependencies)
- `vendor/bin/phpcs inc/checkout/class-checkout.php tests/unit/Checkout_Step_Fields_Test.php` — passes
- `node_modules/.bin/eslint assets/js/checkout.js --ignore-pattern '*.min.js'` — passes (Browserslist age warning only)
- `git diff --check` — passes

Resolves #1332

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced checkout validation to more accurately handle multi-step forms, reducing blocking errors caused by fields previously submitted on earlier steps that are no longer visible on the current step.

* **Tests**
  * Expanded test coverage for multi-step checkout field validation and step-specific field partitioning.

### DIFF
--- a/assets/js/checkout.js
+++ b/assets/js/checkout.js
@@ -617,12 +617,27 @@
 					 * username, password on step 4) from blocking submission of
 					 * earlier steps that do not include those fields.
 					 *
-					 * Falls back to all rules when step_fields is unavailable
-					 * (legacy single-step forms).
+					 * Falls back to the rendered form fields when step_fields is
+					 * unavailable or stale. This keeps multi-step guest checkouts from
+					 * validating account fields that were already submitted on a prior
+					 * step and are no longer present in the DOM.
 					 */
 					const stepFields = (typeof wu_checkout !== 'undefined' && wu_checkout.step_fields) ? wu_checkout.step_fields : null;
 					const currentStep = jQuery('input[name="checkout_step"]').val();
+					const renderedFields = [];
 					let rules = allRules;
+
+					jQuery('#wu_form').serializeArray().forEach(function (field) {
+
+						const fieldName = field.name.replace(/\[\]$/, '');
+
+						if (allRules[ fieldName ] && renderedFields.indexOf(fieldName) === -1) {
+
+							renderedFields.push(fieldName);
+
+						}
+
+					});
 
 					if (stepFields && currentStep && stepFields[ currentStep ]) {
 
@@ -630,13 +645,23 @@
 
 						rules = {};
 
-						allowedFields.forEach(function(fieldId) {
+						allowedFields.forEach(function (fieldId) {
 
 							if (allRules[ fieldId ]) {
 
 								rules[ fieldId ] = allRules[ fieldId ];
 
 							}
+
+						});
+
+					} else if (currentStep && renderedFields.length) {
+
+						rules = {};
+
+						renderedFields.forEach(function (fieldId) {
+
+							rules[ fieldId ] = allRules[ fieldId ];
 
 						});
 
--- a/inc/checkout/class-checkout.php
+++ b/inc/checkout/class-checkout.php
@@ -2324,9 +2324,23 @@ class Checkout {
 		if ($this->checkout_form) {
 			foreach ($this->checkout_form->get_steps_to_show() as $step) {
 				if ( ! empty($step['id']) && ! empty($step['fields'])) {
-					$step_fields[ $step['id'] ] = array_values(
-						array_filter(array_column($step['fields'], 'id'))
-					);
+					$field_ids = [];
+
+					foreach ($step['fields'] as $field) {
+						if ( ! empty($field['id'])) {
+							$field_ids[] = $field['id'];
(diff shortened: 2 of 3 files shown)


superdav42:
MERGE_SUMMARY

Updated `fix/checkout-validation-1332` onto current `main` and pushed the rebased checkout validation scoping fix.

Evidence:
- `bash bin/check-env.sh` — passed; vendor, npm, WP-CLI, shared WordPress install, and `/tmp/wordpress-tests-lib` available.
- `WP_TESTS_DIR=/tmp/wordpress-tests-lib vendor/bin/phpunit --filter Checkout_Step_Fields_Test` — passed (2 tests, 31 assertions; existing dependency deprecation notices only).
- `vendor/bin/phpcs inc/checkout/class-checkout.php tests/unit/Checkout_Step_Fields_Test.php` — passed.
- `node_modules/.bin/eslint assets/js/checkout.js --ignore-pattern '*.min.js'` — passed (Browserslist age warning only).
- `git diff --check` — passed.

Branch evidence:
- Rebased PR commit: `9ab69ff2 fix(checkout): harden step validation scoping`.
- PR diff remains limited to `assets/js/checkout.js`, `inc/checkout/class-checkout.php`, and `tests/unit/Checkout_Step_Fields_Test.php`.

---

PR #1624, opened 2026-07-11
fix(sso): cross-domain SSO redirect to hard-coded /wp/wp-admin/ creates a 404 loop that gets customers IP-banned

kenedytorcatt:
## What happened (real production incident, 2026-07-10)

A brand-new customer on our network simply **visited her own mapped site while logged in**. The SSO flow redirected her to `https://hersite.com/wp/wp-admin/` - a path that only exists on Bedrock-style installs. On a standard install it 404s, and the SSO snippet rendered on the 404 page restarts the flow, appending one more `/wp/wp-admin/` per round:

```
hersite.com/wp/wp-admin/            -> 404
hersite.com/wp/wp-admin/wp/wp-admin/ -> 404
... (verified in access logs: path stacked x7, redirect_to x8)
```

She hit 10 404s in 23 seconds, **Defender's 404 lockout banned her IP for an hour**, and she cancelled her subscription 20 minutes after signing up. The site owner (network admin) got banned the same way while investigating.

## The bug

`SSO::get_sso_redirect_to()` (inc/sso/class-sso.php:1139):

```php
if ( ! empty($return_url) && $this->is_cross_domain_url($return_url) ) {
    return esc_url_raw(rtrim($return_url, '/') . '/wp/wp-admin/');
}
```

The hard-coded `/wp/wp-admin/` suffix assumes a Bedrock directory layout that standard WordPress installs don't have.

## The fix

Return the user to **where they were** (the cleaned `return_url`). Explicit admin intent still travels through the `redirect_to` branch above (untouched). Suffixes already stacked by previous versions are collapsed with a small loop, so URLs that are mid-flight when the fix ships self-heal instead of 404ing one last time.

1 file, +18/-1, validated with 4/4 unit-style tests via Reflection against a staging network plus the production incident URLs.

## Status on our side

Running patched in production (kursopro.com network) since 2026-07-10 - zero SSO 404 loops since. Given that any security plugin with a 404 lockout turns this loop into a **ban of your own paying customers**, we'd appreciate an urgent look. Related money-path fixes in the WooCommerce addon: Ultimate-Multisite/ultimate-multisite-woocommerce#118 and Ultimate-Multisite/ultimate-multisite-woocommerce#119.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved single sign-on redirects across domains by correctly normalizing administrator URLs.
  * Prevented duplicate or malformed administrative path segments in redirect destinations.

PR #1636, opened 2026-07-14
fix(sso): admin auth-redirect preempts incoming cookie-less token - infinite customize.php <-> /login/ loop (site customizer down)

kenedytorcatt:
## Real-world impact (kursopro.com, 2026-07-14)

The site **customizer opened through our panel was down for every subsite**. The embedding UI shows a generic error while the iframe silently loops. Access log (~2 rounds per second):

```
GET subsite/wp-admin/customize.php?wu_sso_token=VALID   -> 302
GET main/login/?sso=login&redirect_to=...customize.php  -> 303 (mints ANOTHER token)
GET subsite/wp-admin/customize.php?wu_sso_token=NEW     -> 302
... forever
```

## The bug

`handle_auth_redirect()` kicks off an SSO round-trip for any wp-admin request that is not logged in and not on the main domain. But when the request **already carries a valid `?wu_sso_token=`** (the cookie-less flow), the kick-off fires **before** `handle_cookie_less_sso_token()` (init:4) can consume it - so the token is never consumed, the main site mints another one, and the loop never converges. The `wu_sso_redirect_attempts` breaker only degrades it to a login form inside the iframe, which the embedding UI renders as an error.

Front-end URLs are unaffected (no kick-off there) - we verified a fresh token on `subsite/?wu_sso_token=...` consumes perfectly (auth cookies set), while the same flow on `wp-admin/customize.php` never even reaches the receiver (zero lines in the sso log, not even the already-used error for a burned token).

## The fix

Mirror the existing `$sso_path` short-circuit right above: when `wu_sso_token` is present in the current request, return `true` so the auth redirect is bypassed and the init:4 receiver consumes the token.

1 file, +13 lines (comment-heavy), no behavior change for requests without a token.

## Verified end-to-end in production

After deploying this patch: `customize.php?wu_sso_token=FRESH` -> 302 to the clean URL **with auth cookies set** -> following with cookies -> **HTTP 200, customizer opens**.

Related: #1624 (merged - thanks for the fast turnaround!), Ultimate-Multisite/ultimate-multisite-woocommerce#118, Ultimate-Multisite/ultimate-multisite-woocommerce#119.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue that could cause repeated SSO redirects when signing in without cookies.
  * Cookie-less SSO tokens are now processed correctly before starting another authentication flow.

### DIFF
--- a/inc/sso/class-sso.php
+++ b/inc/sso/class-sso.php
@@ -390,6 +390,19 @@ class SSO {
 		if ($this->input($sso_path) && $this->input($sso_path) !== 'done') {
 			return true;
 		}
+		/*
+		 * A request that already carries a cookie-less SSO token must not kick
+		 * off a new SSO round-trip: handle_cookie_less_sso_token() (init:4) is
+		 * about to consume it and set the auth cookies. Without this
+		 * short-circuit, wp-admin URLs like customize.php?wu_sso_token=... are
+		 * redirected to the main /login/ BEFORE consumption, the main site mints
+		 * yet another token and the browser loops forever between customize.php
+		 * and /login/ (~2 rounds/sec observed in production access logs). This
+		 * mirrors the $sso_path short-circuit right above.
+		 */
+		if ('' !== (string) $this->input('wu_sso_token', '')) {
+			return true;
+		}
 
 		$should_skip_redirect = $this->get_isset($_COOKIE, 'wu_sso_denied', false);
 


superdav42 (approved):

superdav42:
DISPATCH_CLAIM nonce=6aaf12f962d62aa505ce82d005d64edf runner=superdav42 ts=2026-07-16T12:15:34Z max_age_s=120 version=3.32.131 opencode_version=1.18.2 lease_token=6aaf12f962d62aa505ce82d005d64edf device=device-1783824528-2609248-26808 session=issue-1636 phase=prelaunch expires_at=1784204255

superdav42:
REVIEW_FOLLOWUP_CREATED source_pr=1636 issue=1639 fingerprint=source-pr-1636 runner=superdav42 ts=2026-07-16T12:15:56Z

PR #1641, opened 2026-07-16
Fix shortcode generator warnings and late add-on forms

vuckro:
## Summary

- support array-based `required` values in shortcode generator visibility rules
- keep conditionally displayed attributes in the generated shortcode when any allowed value matches
- register generator/customizer forms immediately when add-on elements initialize after `plugins_loaded`

## Reproduction

1. Open Network Admin → Ultimate Multisite → Shortcodes.
2. Generate a **Magic Link URL** shortcode.
3. The `link_text` and `open_in_new_tab` fields each trigger `Array to string conversion` from `inc/ui/class-base-element.php`.
4. With Ultimate Multisite: Emails 1.1.3 active, generate **Email Management**.
5. The modal displays **Form not available** because the add-on creates its element during `plugins_loaded`, after the core registration callback can run.

## Validation

- reproduced both PHP warnings before the change and zero after it
- confirmed `shortcode_email-management` is missing before the change and registered after it
- PHP_CodeSniffer passes for changed PHP files
- ESLint passes for `assets/js/vue-apps.js`
- regenerated `assets/js/vue-apps.min.js`
- added regression coverage for array requirements and late element form registration

PHPUnit could not start locally because the WordPress test library is not installed in this checkout.

Reproduced with WordPress 7.0.1, PHP 8.2.29, Ultimate Multisite 2.14.1, and Ultimate Multisite: Emails 1.1.3.

@superdav42

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved conditional field visibility for requirements with multiple accepted values.
  - Fixed form registration for elements initialized after application plugins have loaded.
  - Prevented warnings when rendering generator modals with array-based field requirements.

- **Tests**
  - Added coverage for late-loaded form registration and multi-value requirement handling.


vuckro:
@superdav42 The fix has now also been validated manually on the affected local multisite: Magic Link URL generates without PHP warnings, and the Email Management generator opens normally instead of showing “Form not available”. The PR is ready for review.

superdav42 (approved):
Reviewed the shortcode generator and late form registration changes.

- Base_Element::init() now handles elements initialized after plugins_loaded without waiting for a hook that has already fired.
- Array-based shortcode requirements are handled in PHP-generated visibility expressions and in the Vue shortcode builder.
- Removed the generated assets/js/vue-apps.min.js delta so the feature branch only carries source/test changes.
- CI is green across code quality, build, Cypress, PHP 8.2/8.3/8.4/8.5, lint, and performance.

superdav42:
DISPATCH_CLAIM nonce=d1ccec3d3a9d19b1843ef8c5253ab786 runner=superdav42 ts=2026-07-17T12:57:38Z max_age_s=120 version=3.32.145 opencode_version=1.18.3 lease_token=d1ccec3d3a9d19b1843ef8c5253ab786 device=device-1783824528-2609248-26808 session=issue-1641 phase=prelaunch expires_at=1784293180

PR #1643, opened 2026-07-16
Restore add-on product types in the product editor

vuckro:
## Summary

- restore add-on product types in the product editor dropdown
- preserve the longer usage descriptions for the four core product types
- isolate `Enum::to_array()` caches by subclass so one enum cannot return another enum's labels
- add regression coverage for an add-on type after another enum has already been loaded

## Root cause

`Product_Edit_Admin_Page::get_product_type_options()` was changed in 2.13.2 to return a hard-coded list of Plan, Package, Service, and Demo. The product type enum remained extensible, but the editor no longer consumed its filtered labels.

There was also an order-dependent issue in `Enum::to_array()`: its inherited static local cached one instance shared by every enum subclass. For example, loading `Payment_Status::to_array()` first caused a later `Product_Type::to_array()` call to use the payment-status hook and labels. Add-on product filters then could not run reliably.

As a result, add-ons could register a valid product type that worked in the model layer while administrators had no way to select it. Ultimate Multisite: Emails 1.1.3 exposes this as a missing **Email** option.

The editor now starts with `Product_Type::to_array()` and replaces only the four core labels with their detailed descriptions. `Enum::to_array()` caches an instance per concrete enum class, so add-on labels remain stable regardless of call order.

## Validation

- PHP syntax and Git diff checks pass locally
- PHP_CodeSniffer passes for the changed product editor production file
- PHPStan passes for the changed product editor production file
- local call-order probe loads payment statuses first, then confirms product options contain `plan` and `email` but not payment status keys
- local WordPress validation returns `plan`, `package`, `service`, `demo`, and `email`
- local product sections include `email_settings`
- configured cPanel integration is available and its connection test succeeds
- transactional add-on test covers secondary membership creation, activation, provisioning, shortcode rendering, and rollback
- regression test covers a type registered by an add-on after another enum has initialized its cache

The local PHPUnit runner could not start because the WordPress test library is not installed in this checkout; the full matrix runs in CI.

Reproduced with WordPress 7.0.1, Ultimate Multisite 2.14.1, and Ultimate Multisite: Emails 1.1.3.

@superdav42

### DIFF
--- a/inc/admin-pages/class-product-edit-admin-page.php
+++ b/inc/admin-pages/class-product-edit-admin-page.php
@@ -695,12 +695,14 @@ class Product_Edit_Admin_Page extends Edit_Admin_Page {
 	 */
 	protected function get_product_type_options() {
 
-		return [
-			Product_Type::PLAN    => __('Plan — Subscription tier that creates and manages customer sites.', 'ultimate-multisite'),
-			Product_Type::PACKAGE => __('Package — Add-on sold with plans to unlock extra features or resources.', 'ultimate-multisite'),
-			Product_Type::SERVICE => __('Service — Work or support customers can buy without creating a site.', 'ultimate-multisite'),
-			Product_Type::DEMO    => __('Demo — Temporary trial site for prospects to test before buying.', 'ultimate-multisite'),
-		];
+		$options = Product_Type::to_array();
+
+		$options[ Product_Type::PLAN ]    = __('Plan — Subscription tier that creates and manages customer sites.', 'ultimate-multisite');
+		$options[ Product_Type::PACKAGE ] = __('Package — Add-on sold with plans to unlock extra features or resources.', 'ultimate-multisite');
+		$options[ Product_Type::SERVICE ] = __('Service — Work or support customers can buy without creating a site.', 'ultimate-multisite');
+		$options[ Product_Type::DEMO ]    = __('Demo — Temporary trial site for prospects to test before buying.', 'ultimate-multisite');
+
+		return $options;
 	}
 
 	/**
--- a/inc/database/engine/class-enum.php
+++ b/inc/database/engine/class-enum.php
@@ -62,7 +62,7 @@ abstract class Enum {
 	 * Returns an array with values => labels.
 	 *
 	 * @since 2.0.0
-	 * @return void
+	 * @return array
 	 */
 	abstract protected function labels();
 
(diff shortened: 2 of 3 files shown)


vuckro:
Companion findings for Ultimate Multisite: Emails 1.1.3:

- `wu_product_type_to_array` should return a descriptive editor label such as **Email — Mailbox plan that lets customers create and manage email accounts.**
- `wu_available_product_type_labels` should continue returning the short **Email** label for badges and lists.
- `allowed_templates` should be included in `Email_Product_Manager::hide_hosting_limit_sections()` because an email product does not create a site; it inherits the domain from the parent site membership.

These companion adjustments are applied and validated on the reporting local installation. The add-on source repository is not accessible from the reporting account, so they are documented here for the add-on follow-up.

@superdav42

vuckro:
Additional end-to-end finding in Ultimate Multisite: Emails 1.1.3:

The Order Bump correctly creates the secondary email membership as `pending`, but there is currently no lifecycle path that activates it when the primary membership becomes active. The resulting customer can own an email product but still have a pending email membership.

Local companion fix validated:

1. Listen to `wu_membership_activated` in `Checkout_Integration`.
2. Read the primary membership's `wu_email_membership_ids`.
3. For each linked membership, verify that its product type is `email` and its current status is `pending`.
4. Set it to `active`, save it, and log any `WP_Error` through the existing `emails-membership` channel.

Transactional integration coverage on the reporting installation confirmed:

- order bump creates and links the secondary email membership;
- primary activation changes the linked email membership from `pending` to `active`;
- site provisioning records the inherited domain and the `cpanel` provider;
- `[wu_email_management]` renders the plan and the **Create Account** action;
- rollback leaves the original database state intact.

The local demonstration state now uses the intended separation: primary membership `#1` (Free) and linked email membership `#4` (Email plan). Site Templates are intentionally hidden for email products because the domain comes from the primary site's membership.

The add-on repository is still not accessible from this account, so this patch is applied locally and documented here for the add-on follow-up.

@superdav42

superdav42 (approved):
Reviewed the product type extensibility fix.

- The product editor now starts from Product_Type::to_array(), so add-on product types remain available while core labels keep their descriptive copy.
- Enum to_array() cache isolation prevents earlier enum calls from applying the wrong hook/labels to product types.
- The regression covers both descriptive add-on labels and cross-enum cache leakage.
- CI is green across code quality, build, Cypress, PHP 8.2/8.3/8.4/8.5, lint, and performance.

PR #1692, opened 2026-08-03
fix(checkout): prevent epoch membership expiration

superdav42:
## Summary

- use the calculated next charge date when recurring billing starts immediately
- preserve null expiration dates for lifetime memberships
- cover persisted, email-placeholder, and customer-facing expiration output

## Root cause

Recurring carts without trials report a billing start timestamp of `0`. Formatting that value directly persisted a January 1970 membership expiration, which then propagated to email event payloads and customer views.

## Verification

- `vendor/bin/phpunit --no-coverage --filter 'WP_Ultimo\\Checkout\\Checkout_Test'` — 204 tests, 349 assertions, 1 skip
- `vendor/bin/phpcs inc/checkout/class-checkout.php tests/WP_Ultimo/Checkout/Checkout_Test.php`
- `vendor/bin/phpstan analyse inc/checkout/class-checkout.php tests/WP_Ultimo/Checkout/Checkout_Test.php --no-progress`
- `git diff --check`

---
[aidevops.sh](https://aidevops.sh) v3.32.219 plugin for [OpenCode](https://opencode.ai) v1.18.9 with gpt-5.6-sol spent 2h 16m and 440,196 tokens on this with the user in an interactive session.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed membership expiration dates for recurring plans without trials.
  - Prevented memberships from incorrectly expiring at the Unix epoch.
  - Improved cleanup of expired pending payments and memberships.
  - Recent pending payments and memberships are no longer cancelled prematurely.
  - Cancelled expired pending memberships now also remove associated payment metadata.

- **Tests**
  - Added regression coverage for recurring memberships, recent pending payments, and expired payment cleanup.

PR #1830, opened 2026-09-09
fix: normalize spaces in generated usernames

superdav42:
## Summary

- Normalize whitespace to dots in `wu_username_from_email()` before checking blocked usernames and existing accounts.
- Match the dotted-username convention already used by `WooCommerce_Subscriptions_Compat::woocommerce_new_customer_username_no_space()`.
- Add regression coverage in `tests/WP_Ultimo/Functions/Customer_Functions_Test.php` for compound first/last names, repeated spaces, unchanged normalized names, collisions, blocked names, and persisted accounts created without a supplied password.

## Why

`sanitize_user(..., true)` preserves internal spaces. Ultimate Multisite's generator sanitizes first/last names and joins them with a dot, so `Alice Beth` / `Example` previously generated `alice beth.example`.

The auto-generated-password checkout path passes an empty password to `wu_create_customer()`, which uses `register_new_user()`. That path can persist the space-containing username. Normalizing the generated candidate before validation makes new names consistent without renaming existing users or changing explicitly supplied usernames.

Normalization must happen before `username_exists()` and `illegal_user_logins` checks: if `alice.beth.example` already exists or is blocked, the generator must handle that normalized name rather than checking a different space-containing name.

This is an independent preventive fix, not an attribution of the original SSO report to Ultimate Multisite's account generation. Related PR #1826 fixes authentication-cookie encoding for existing accounts containing spaces and remains necessary.

## Scope and compatibility

- Only the generated candidate changes; existing accounts are not migrated or renamed.
- Existing suffix generation and the final `wu_username_from_email` extension filter remain intact.
- Includes a targeted PHPCS exception for the pre-existing public `$default` parameter of `wu_get_customer_meta()`. The same warning was reproduced on the unchanged base file; keeping the parameter name avoids breaking PHP named-argument callers. No lint configuration or hook is weakened.

## Verification

Before the production fix, the username-focused run produced **6 expected failures**, including the saved username retaining its space.

After the fix:

- `vendor/bin/phpunit tests/WP_Ultimo/Functions/Customer_Functions_Test.php --no-coverage`: **28 tests, 45 assertions, passing**.
- `vendor/bin/phpcs inc/functions/customer.php tests/WP_Ultimo/Functions/Customer_Functions_Test.php`: passing.
- `vendor/bin/phpstan analyse inc/functions/customer.php tests/WP_Ultimo/Functions/Customer_Functions_Test.php --no-progress`: no errors.
- `php -l inc/functions/customer.php`: passing.
[...]

PR #1837, opened 2026-09-09
fix(sso): preserve SSO for networks upgrading without a saved setting

kenedytorcatt:
## Summary

`#1741` changed the SSO fallback so that **fresh** installations start with SSO off. The
implementation changes `get_setting_defaults()`, and that map is also what an **existing**
network reads when it has no saved `enable_sso` key — so upgrading to 2.15.2 silently turns
SSO off for those networks.

```
inc/class-settings.php:2205  (2.15.1)   'enable_sso' => 1,
inc/class-settings.php:2235  (2.15.2)   'enable_sso' => 0,
```

`Settings::get_setting()` falls back to that map when the key is absent
(`inc/class-settings.php:279`), and nothing persists the previous value on upgrade. The PR
description for #1741 says it preserves *"existing installations' explicitly saved SSO
setting"* — which is exactly the set that was never at risk. The set that *is* at risk is the
one with **no** saved value.

**Who is affected:** a network whose saved settings array predates the `enable_sso` key. A
network that ran the installer after the key existed already has it persisted
(`inc/installers/class-default-content-installer.php:271` saves the whole defaults map), so
most installs are unaffected. The failure mode is what makes it worth fixing: SSO stops
working across mapped domains, the settings page shows the toggle off as if that had always
been the case, and nothing points at the upgrade.

## The change

`Settings::maybe_preserve_legacy_sso_default()` on `init` at priority 1: if the network has
settings saved but no `enable_sso` key, it writes the pre-2.15.2 value once.

- A **fresh** installation is untouched — the installer persists `enable_sso => 0` before this
  ever runs, so the `isset()` short-circuits and the new default stands.
- A network that **explicitly disabled** SSO keeps its choice.
- **No migration flag and no version marker.** The saved key is its own guard: once written,
  every later request returns on the `isset()`. The only cost in steady state is the settings
  array the plugin already loads.

## Verification

- `php -l` on both changed files.
- 4 unit tests added to `tests/WP_Ultimo/Settings_Test.php`: the value is restored for a
  network without one, a saved `0` is never overwritten, a fresh install is not seeded, and the
  migration writes only once (a later "disable SSO" survives a second run).
- I could not run `phpcs` / `phpstan` / `phpunit` locally (no `vendor/` in my checkout), so CI
  is the first full run.

## Context

We run 2.15.2 on a live 23-site network (upgraded 2026-09-09, staging first, then production)
and validated it end to end: a full signup creates the membership as `trialing` with the
expiration at exactly +15 days — the epoch expiration from #1692 is gone — the pending site is
published after the checkout transaction commits, third-party hooks at priority 9999 still
fire, and there were no fatals after the swap. Our own network was unaffected by this SSO issue
because `enable_sso` was explicitly saved; we found it while diffing 2.15.1 against 2.15.2
before rolling the update out.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved the previously enabled SSO default for existing networks without a saved SSO setting.
  * Existing saved SSO preferences remain unchanged.
  * New installations continue using the current default behavior.
  * The compatibility setting is applied only once.

### DIFF
--- a/inc/class-settings.php
+++ b/inc/class-settings.php
@@ -74,6 +74,8 @@ class Settings implements \WP_Ultimo\Interfaces\Singleton {
 	 */
 	public function init(): void {
 
+		add_action('init', [$this, 'maybe_preserve_legacy_sso_default'], 1);
+
 		add_action('init', [$this, 'handle_legacy_filters'], 2);
 
 		add_action('wu_render_settings', [$this, 'handle_legacy_scripts']);
@@ -90,6 +92,43 @@ class Settings implements \WP_Ultimo\Interfaces\Singleton {
 		// plugin deactivation in the correct state.
 		add_action('wu_activation', [$this, 'sync_wp_registration_option']);
 		add_action('wu_after_save_settings', [$this, 'sync_wp_registration_option']);
+	}
+
+	/**
+	 * Keep SSO on for networks that were relying on the pre-2.15.2 fallback.
+	 *
+	 * `get_setting_defaults()` returned `enable_sso => 1` until 2.15.1, so a network that
+	 * never opened the settings page still had SSO on. 2.15.2 changed that fallback to `0`
+	 * to give fresh installations a safer default, but the fallback is also what an older
+	 * network without a saved `enable_sso` key reads, so upgrading silently turns SSO off
+	 * for it — with no notice and nothing in the settings UI to explain the change.
+	 *
+	 * This writes the previous default once, and only for a network that is already set up
+	 * and has no value of its own. A fresh installation is untouched: the installer persists
+	 * the whole defaults map, so `enable_sso` is already present (and `0`) by the time this
+	 * runs. The saved key is its own guard — once written, this returns on the `isset()`
+	 * below and never writes again, so there is no migration flag to store or read.
+	 *
+	 * @since 2.15.3
+	 * @return void
+	 */
+	public function maybe_preserve_legacy_sso_default(): void {
+
+		$settings = $this->get_all();
+
+		/*
+		 * Nothing saved yet: the plugin has not been set up on this network, and the
+		 * installer will persist the current defaults, including the new SSO one.
+		 */
+		if (empty($settings)) {
+			return;
+		}
+
+		if (isset($settings['enable_sso'])) {
+			return;
+		}
+
+		$this->save_setting('enable_sso', 1);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


superdav42 (approved):

superdav42:
Kenedy, thank you for catching this regression and for providing such a careful explanation and focused tests. I’m sorry for the inconvenience caused by our 2.15.2 SSO-default change, and for the extra work it created for you.

I reviewed the migration guard and test coverage and found no blocking issues. It correctly preserves the legacy behavior only when an existing settings array lacks enable_sso, while retaining explicit choices and the safer default for fresh installations. We’ll merge this once the remaining required CI checks pass.



---
[aidevops.sh](https://aidevops.sh) v3.32.317 plugin for [OpenCode](https://opencode.ai) v1.18.30 with gpt-5.6-sol

kenedytorcatt:
Makes sense — thanks for the context, and the username whitespace work (#1830 / #1826) looks like it removes the actual cause of the loop.

To be clear about what this PR is and isn't: it takes no position on the default itself, only on the upgrade being **silent**. A network that has SSO working today updates, it stops working, and the settings toggle shows "off" as if it had always been that way — that's the afternoon someone loses.

So:

- If your intent is off-for-everyone, including networks that already had it on, **close this** — no argument from me.
- If you'd rather keep your new default and just remove the silence, I'm happy to rework it into a notice instead: on a network that had no saved `enable_sso` value, show a dismissible network-admin notice saying SSO is now off by default, why, and where the toggle is. Same protection against the lost afternoon, zero change to your default.

Either way is fine — just say which and I'll adjust or close it myself.

For what it's worth, we weren't bitten by this one: our `enable_sso` was explicitly saved, and we have 0 of 191 users with whitespace in `user_login`, so neither issue reached us. We found it diffing 2.15.1 against 2.15.2 before rolling the update out to a live 23-site network.

superdav42:
DISPATCH_CLAIM nonce=4caf7285bd6999679e187287c796158c runner=superdav42 ts=2026-09-10T13:19:42Z max_age_s=120 version=3.32.317 opencode_version=1.18.30 lease_token=4caf7285bd6999679e187287c796158c device=device-1783824528-2609248-26808 session=issue-1837 phase=prelaunch expires_at=1789046504

superdav42:
REVIEW_FOLLOWUP_CREATED source_pr=1837 issue=1846 fingerprint=source-pr-1837 runner=superdav42 ts=2026-09-10T13:20:07Z

PR #1857, opened 2026-09-11
GH#1853: fix(site-duplicator): refresh cloned site roles before hooks

### LINKED ISSUE #1853: Cloned site can come up with a single role in user_roles, locking the owner out of their own dashboard
A site created through the WooCommerce checkout came up with **one** role in `wp_<id>_options.user_roles` instead of the usual 12. Its owner held the role `administrator`, which no longer existed on that blog, so the customer got *"Access denied"* on their own dashboard minutes after paying.

I could not reproduce it on demand, so this is a report with a signature and a hypothesis rather than a pull request. Everything below is measured on the affected network.

## What the broken value looked like

`user_roles` did not contain a damaged or truncated map. It contained **one complete role**: `tutor_instructor`, with all 36 of its capabilities and a `name` already translated into the site language.

That shape is the interesting part. A corrupted copy, a partial write or a failed `INSERT ... SELECT` would leave a mangled or empty map. A map holding exactly one fully-formed role is what `WP_Roles::add_role()` produces when it runs against a `WP_Roles` instance whose `roles` array is **empty**: the whole option is then persisted with that single entry. In other words it looks like a third-party plugin (Tutor LMS, in our case) creating its own role while the new blog's role map was not populated in memory.

The translated `name` says it happened inside a full web request with plugins and text domains loaded, not in CLI.

Everything else about the clone was correct: 19 tables, 1603 options, 77 pages, all copied fine.

## Why I suspect the duplicator's window

Reading `inc/duplication/data.php` and `inc/helpers/class-site-duplicator.php` on `main`:

- `class-site-duplicator.php:397` — `wpmu_create_blog()` creates the blog, and WordPress populates its 12 roles at that point.
- `data.php:129` — `copy_data()` then runs `DROP TABLE IF EXISTS` on the destination tables, including `wp_<id>_options`, followed by `SHOW CREATE TABLE` + `INSERT ... SELECT` from the template. The blog's options table therefore **does not exist** for a window, and all of this happens in raw SQL, outside the options API.
- `class-site-duplicator.php:552` — `wp_cache_flush()`.
- `class-site-duplicator.php:583` — `do_action('wu_duplicate_site', ...)`.

The part that made me open this: `grep -rn 'wp_roles()\|WP_Roles\|for_site(' inc/helpers/class-site-duplicator.php inc/duplication/` returns **0 matches**. The duplicator flushes the object cache but never re-seeds the in-memory `$GLOBALS['wp_roles']` singleton for the destination blog. Any `WP_Roles` instantiated by a third party inside that window would hold an empty map in memory, and a later `add_role()` would persist the option with just that one role — which is exactly the signature above.

We run a **persistent object cache (Redis)**, which is relevant here because it changes how long a stale in-memory or cached role map can survive.

## Environment

| | |
|---|---|
| Ultimate Multisite | 2.15.2 |
| WordPress | 6.9 multisite |
| WooCommerce / Subscriptions | 11.0.0 / 9.2.0 |
| PHP | 8.4 |
| Object cache | persistent (Redis) |
| Signup path | WooCommerce checkout **with a coupon** (not the `free` gateway) |
| Template / destination | blog 157 → blog 388 |
| Date | 2026-09-09, the same day 2.15.2's `Membership_Manager::commit_checkout_transaction` started publishing the pending site only after the transaction is confirmed |

## What I could NOT reproduce

Two signups on staging with **2.16.0** produced a healthy site both times: 12 roles, `administrator` with 293 capabilities. But both went through the **`free`** gateway, not WooCommerce with a coupon, so they do not exercise the same path and I am not claiming 2.16.0 fixed it.

If you want to try it, the combination that produced it here was: WooCommerce checkout with a 100%-off coupon, a template whose site has a plugin that registers its own role on init, and a persistent object cache.

## The question

Is it guaranteed that no in-memory `WP_Roles` instance survives the `DROP TABLE` / `INSERT ... SELECT` window in `MUCD_Data::copy_data()`, given that `wp_cache_flush()` does not rebuild `$GLOBALS['wp_roles']`?

If it is not guaranteed, a `wp_roles()->for_site($to_site_id)` next to the `wp_cache_flush()` at `class-site-duplicator.php:552`, before the `do_action('wu_duplicate_site')` at 583, would close it cheaply. I am happy to prepare that as a PR if you think the reasoning holds, but I did not want to send a fix for something I cannot make fail on demand.

## Recovery, for anyone who hits this

Copying `user_roles` from the template blog restores the site (12 roles, `administrator` with 293 capabilities). The owner's membership of the new blog may also need restoring: in our case the author of a cloned course lost their membership of the new blog as a side effect.

superdav42:
## Summary

Refresh the cloned site's in-memory WordPress role map after copying its options table and before extension hooks run, preventing hook-added roles from replacing the copied map.

## Files Changed

inc/helpers/class-site-duplicator.php, tests/WP_Ultimo/Helpers/Site_Duplicator_Test.php

## Runtime Testing

- **Risk level:** Medium
- **Verification:** self-assessed — vendor/bin/phpunit --filter test_wu_duplicate_site_action_receives_reloaded_roles; vendor/bin/phpcs inc/helpers/class-site-duplicator.php tests/WP_Ultimo/Helpers/Site_Duplicator_Test.php; vendor/bin/phpstan analyse --no-progress --error-format=table inc/helpers/class-site-duplicator.php

Resolves #1853

---
[aidevops.sh](https://aidevops.sh) v3.32.317 plugin for [OpenCode](https://opencode.ai) v1.18.30 with gpt-5.6-terra spent 8m and 124,959 tokens on this as a headless worker.

## Summary by CodeRabbit

* **Bug Fixes**
  * Site duplication now correctly reloads WordPress roles before duplication actions run.
  * Roles created on the template site are available on the duplicated site during duplication hooks.

* **Tests**
  * Added regression coverage to verify role availability during the site duplication process.

### DIFF
--- a/inc/helpers/class-site-duplicator.php
+++ b/inc/helpers/class-site-duplicator.php
@@ -550,6 +550,15 @@ class Site_Duplicator {
 
 		$profile_stage = microtime(true);
 		wp_cache_flush();
+
+		/*
+		 * Rebuild the in-memory role map from the copied options table before
+		 * invoking extension hooks. A cache flush does not reset WP_Roles, and
+		 * a hook that adds a role with an empty role map would overwrite
+		 * user_roles with that single role.
+		 */
+		wp_roles()->for_site($args->to_site_id);
+
 		self::profile_sovereign_provisioning_stage(
 			(int) $args->to_site_id,
 			'um_duplicator.wp_cache_flush',
--- a/tests/WP_Ultimo/Helpers/Site_Duplicator_Test.php
+++ b/tests/WP_Ultimo/Helpers/Site_Duplicator_Test.php
@@ -822,6 +822,45 @@ class Site_Duplicator_Test extends WP_UnitTestCase {
 	}
 
 	/**
+	 * Test duplicated site roles are available to extension hooks.
+	 */
+	public function test_wu_duplicate_site_action_receives_reloaded_roles() {
+		$role_slug    = 'site_duplicator_template_role';
+		$role_at_hook = false;
+		$result       = false;
+		$role_hook    = function () use (&$role_at_hook, $role_slug) {
+			$role_at_hook = wp_roles()->get_role($role_slug);
+		};
+
+		switch_to_blog($this->template_site_id);
+		add_role($role_slug, 'Site Duplicator Template Role', ['read' => true]);
+		restore_current_blog();
+
+		add_action('wu_duplicate_site', $role_hook);
+
+		try {
+			$result = Site_Duplicator::duplicate_site(
+				$this->template_site_id,
+				'Role Hook Site',
+				[
+					'domain' => 'role-hook.example.com',
+					'path'   => '/',
+					'title'  => 'Role Hook Site',
+				]
+			);
+
+			$this->assertIsInt($result);
+			$this->assertNotFalse($role_at_hook);
+		} finally {
+			remove_action('wu_duplicate_site', $role_hook);
+
+			if ($result) {
+				wpmu_delete_blog($result, true);
+			}
+		}
+	}
+
+	/**
 	 * Test backfill_kit_settings is a no-op when template has no Kit.
 	 */
 	public function test_backfill_kit_settings_noop_without_elementor() {


superdav42:
## Completion Summary

- **What**: Refresh the cloned site's in-memory WordPress role map after copying its options table and before extension hooks run, preventing hook-added roles from replacing the copied map.
- **Issue**: #1853
- **Files changed**: inc/helpers/class-site-duplicator.php, tests/WP_Ultimo/Helpers/Site_Duplicator_Test.php
- **Testing**: vendor/bin/phpunit --filter test_wu_duplicate_site_action_receives_reloaded_roles; vendor/bin/phpcs inc/helpers/class-site-duplicator.php tests/WP_Ultimo/Helpers/Site_Duplicator_Test.php; vendor/bin/phpstan analyse --no-progress --error-format=table inc/helpers/class-site-duplicator.php
- **Key decisions**: none



---
[aidevops.sh](https://aidevops.sh) v3.32.317 plugin for [OpenCode](https://opencode.ai) v1.18.30 with gpt-5.6-terra spent 9m and 124,959 tokens on this as a headless worker.

---
Merged via PR #1857 to main.

